PR #11, opened 2023-07-28
feat: setting up base

harshtandiya:
Set up the base html. I have extended the already established base of Frappe and built upon it. Figured out a way to add custom CSS and Scripts. This should serve as a good enough template. 

- A person can add style to any components by writing CSS in the `styles.css` file in `public/css` folder. Else, they can also create their own and overwrite it using the `{% block page_style %}` block.
- Same for the Javascript. use `common_functions.js` in `public/js` folder. Or, add your own by overwriting `{% block page_script %}` block.


fossunited/templates/foss_base.html
```diff
@@ -0,0 +1,6 @@
+{% extends "templates/base.html" %}
+
+
+{% block content %}
+    
+{% endblock %}
```
NagariaHussain: Where are you including the css link?

Also, define the HTML here, inheriting from base.html here will does not make sense. Write the outer HTML and define blocks for pages where you will use this. 

Read Jinja docs on template extending to get a better idea...

NagariaHussain (changes requested):

harshtandiya:
@NagariaHussain Can we merge this one?

PR #15, opened 2023-08-18
feat: establish chapter form live

harshtandiya:
Establish a Chapter Form Added.

- The title is dynamic. Changes according to the chapter type selected. If the chapter type in form is City Chapter, it will show "Establish a City Chapter", whereas if it is college chapter, title will be "Establish a FOSS Club".
- The idea is to pass the chapter type via the URL. 

image

image


NagariaHussain:
SS?

harshtandiya:
Updated @NagariaHussain

NagariaHussain (approved):

PR #57, opened 2023-11-21
chore: quick minor fixes

arunmathaisk:
Was just checking out this repo, exploring the doctype organization.

- Unique contatraint on email.

- Institute name field added.

### DIFF
--- a/fossunited/fossunited/doctype/foss_user_profile/foss_user_profile.json
+++ b/fossunited/fossunited/doctype/foss_user_profile/foss_user_profile.json
@@ -60,7 +60,8 @@
    "fieldname": "email",
    "fieldtype": "Data",
    "label": "Email",
-   "options": "Email"
+   "options": "Email",
+   "unique": 1
   },
   {
    "fieldname": "employment_status_section",
@@ -155,7 +156,7 @@
  "image_field": "profile_photo",
  "index_web_pages_for_search": 1,
  "links": [],
- "modified": "2023-10-22 02:27:11.424128",
+ "modified": "2023-11-21 13:13:13.838556",
  "modified_by": "Administrator",
  "module": "FOSSUnited",
  "name": "FOSS User Profile",
--- a/fossunited/fossunited/doctype/institute/institute.json
+++ b/fossunited/fossunited/doctype/institute/institute.json
@@ -10,7 +10,8 @@
  "field_order": [
   "state",
   "city",
-  "country"
+  "country",
+  "institute_name"
  ],
  "fields": [
   {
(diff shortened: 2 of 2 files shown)


fossunited/fossunited/doctype/institute/institute.json
```diff
@@ -35,11 +36,17 @@
    "fieldtype": "Link",
    "label": "Country",
    "options": "Country"
+  },
+  {
+   "fieldname": "institute_name",
+   "fieldtype": "Data",
+   "label": "Name",
+   "reqd": 1
```
harshtandiya, line 44: We do not need to explicitly have a field called `institute_name`.  The naming series for this doctype is done via a prompt to the user. So, Frappe provides a `name` field automatically (which is also a required field) at the time of creation!

fossunited/fossunited/doctype/foss_user_profile/foss_user_profile.json
```diff
@@ -60,7 +60,8 @@
    "fieldname": "email",
    "fieldtype": "Data",
    "label": "Email",
-   "options": "Email"
+   "options": "Email",
+   "unique": 1
```
harshtandiya, line 64: The fields `email` and `full_name` are being fetched from the "User" core doctype of Frappe Framework. So it is not at all needed for it to be unique here.

The idea is when a person signs up for the platform, a core "User" doctype will be created and a seperate FOSS User Profile will be created related to this core doctype. While "User" doctype can handle all the permissions/roles/password related stuff of user, FOSS User Profile will be used to carry the additional info about the user (such as Experience etc.)

fossunited/fossunited/doctype/institute/institute.json
arunmathaisk: Fair enough

fossunited/fossunited/doctype/foss_user_profile/foss_user_profile.json
arunmathaisk: @Mangeshrex I have conveyed the same him. There are a few redundant fields, eg: the reviewers doctype.
UserProfile should ideally be the only single doctype where all suerdata and metadata must be fetched

arunmathaisk: I did not see any validation, Doc Event to auto fetch the same. Maybe I am lacking context regarding your control flow.
Will wait for a bit :)

harshtandiya: Most fields that may seem redundant are there only because the data is being fetched into them from the User Doctype.  Maybe I can make then read-only so that people do not get confused in the future. 

The reason you can't see the validations is probably becauseit is being handled through Frappe Desk. For Ex: The email field in the FOSS User Profile has this:
image
 
This let's the field to autofetch the email value from the User link field every time a doctype element is created/saved.

PR #259, opened 2024-04-26
fix: ui/ux

NagariaHussain:
1. Use `Avatar` component from FrappeUI
2. Proper name for event card component (`EventCardDashboard` -> `EventCard`)
3. Fix layout shift on hover:

  https://github.com/fossunited/fossunited/assets/34810212/87ca4c23-582d-4c8e-8f0c-de33595211d1

4. Some copy

Before:

CleanShot 2024-04-26 at 14 34 44@2x

After: 

CleanShot 2024-04-26 at 14 31 38@2x


NagariaHussain:
Some more things that could be improved:

1. Merge Insights and Status tab in a tab called "Details"
2. Publish/Unpublish form should be action in the header
3. Replace the event and chapter card grids with frappe-ui list views

PR #401, opened 2024-06-19
feat(tickets): add ticket transfer doctype

### LINKED ISSUE #400: [FEATURE] Automate Ticket Transfers
### Feature description

We need to automate ticket transfer process. Right now we need a dedicated person who looks into transfer queries and caters to them individually, it is very _inefficient_

Proposed solution: 

I am going with the assumption that we let the people do the monetary transactions with each other without FOSS United having any part in it. The organization's role will only be to change the ticket details

- Let's have a portal where people can transfer tickets
- People can enter their ticket ID and email + email of the person they want the ticket to be transferred
- Validation of Ticket ID + email. Once validated, email will be sent to the original ticket holder who will approve the request from email.
- Upon approval, the ticket details will be updated and a new email will be sent to the person receiving the ticket.

harshtandiya:
- [x] 🍕Feature

## Description

Automate the ticket transfer process. More details at #400 

## Related Issues & Docs
closes #400 

## Screenshots
image
image

### DIFF
--- a/dashboard/src/pages/TicketTransfer.vue
+++ b/dashboard/src/pages/TicketTransfer.vue
@@ -1,0 +1,255 @@
+<template>
+  <Header />
+  <div class="w-full flex items-center justify-center">
+    <TransferSuccess v-if="inSuccess" />
+    <div v-else class="max-w-screen-xl w-full p-5">
+      <div class="mt-8">
+        <div class="prose">
+          <div class="text-base uppercase font-medium text-gray-600 mb-2">
+            Portal
+          </div>
+          <h1 class="m-0">Ticket Transfers</h1>
+          <div
+            v-if="transfer_settings.doc"
+            v-html="transfer_settings.doc.web_page_description"
+          ></div>
+        </div>
+      </div>
+      <div class="bg-white p-6 rounded border w-full">
+        <div class="flex flex-col md:grid md:grid-cols-2 gap-4">
+          <div class="col-span-2 flex flex-col gap-2">
+            <div class="prose">
+              <h4>Ticket Details</h4>
+            </div>
+            <div class="w-full flex flex-col md:grid md:grid-cols-2 gap-4">
+              <FormControl
+                size="sm"
+                label="Ticket ID  &ast;"
+                v-model="ticketId"
+                type="text"
+                @focusout="
+                  () => {
+                    isTicketValid()
+                  }
+                "
+              />
+              <div class="hidden md:block"></div>
+              <div
+                v-if="ticket.data && event.data"
+                class="col-span-2 grid grid-cols-1 md:grid-cols-2 gap-4"
+              >
+                <FormControl
+                  label="Event"
+                  v-model="event.data.event_name"
+                  type="text"
+                  :disabled="true"
+                />
+                <FormControl
+                  label="Ticket Tier"
+                  v-model="ticket.data.tier"
+                  type="text"
+                  :disabled="true"
+                />
+              </div>
+            </div>
+            <ErrorMessage :message="ticketValidateError" />
+          </div>
+          <div class="col-span-2">
+            <hr />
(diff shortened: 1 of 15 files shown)


dashboard/src/pages/TicketTransfer.vue
```diff
@@ -0,0 +1,255 @@
...
+          <FormControl
+            v-model="reciever.reciever_email"
+            label="Reciever Email &ast;"
+            type="email"
+            size="sm"
+            placeholder="john@fossunited.org"
+          />
+          <FormControl
+            v-model="reciever.reciever_name"
+            label="Reciever Name &ast;"
```
RiyaMathew-11: Typo in this - Reciever -> Receiver 

Needs to be changed in other places too.

RiyaMathew-11 (changes requested):

dashboard/src/pages/TicketTransfer.vue
```diff
@@ -0,0 +1,255 @@
...
+
+const ticketValidateError = ref('')
+const ticketErrors = ref('')
+
+const ticketId = ref('')
+const reciever = reactive({
+  reciever_email: '',
+  reciever_name: '',
+  designation: '',
+  organization: '',
```
RiyaMathew-11, line 155: meta comment - if the purchased ticket has a tshirt alongside - then that would need to be transferred too right? 

How would the case be for different tshirt sizes for transferer and receiver ?

RiyaMathew-11:
Can you also illustrate the case of how transfer approval and rejection status comes in ? 

When does a person see that message? - I am assuming that there is some time gap that would take in the approval process - how is a person able to track his/her ticket transfer status?

Is it in the format of a portal page that the transferer can access ?

dashboard/src/pages/TicketTransfer.vue
harshtandiya: Yes the tshirt will be transferred too.

I can add something for different tshirt size, but will that be okay? I hope there are no problems from logistic side of things

dashboard/src/pages/TicketTransfer.vue
harshtandiya: Outdated commit. Changed in later ones

harshtandiya:
@RiyaMathew-11 

Notification to the owner on ticket transfer, where they can approve or reject.

On Approval -> redirect to website which runs script and changes the status of ticket transfer to completed

Another Notification -> Email to the new owner + old owner that ticket was transferred, also email to new owner with a ticket.pdf

dashboard/src/pages/TicketTransfer.vue
RiyaMathew-11: Well if incorrect tshirt sizes - I think that would also not work? Let's check with Ansh/Vishal on this too.

RiyaMathew-11:
On second thoughts, I feel a little additional step here - an additional email notification for approval. 

The original owner transfers the ticket (only the original owner would have ticket ID - that's like a security key too)- after the process is initiated, why would he/she have to approve/reject it again via CTA from an email?  

I think it would be more easier if - once the `initiate transfer` button is checked, a floating window appears to confirm the transfer again - with Approve/Reject CTA and close the process there ? It reduces one set of emails. 

cc: @rahulporuri @NagariaHussain would like your thoughts too.

harshtandiya:
@RiyaMathew-11 I think it's always better to be on the safer side. We should keep the email approval workflow.

dashboard/src/pages/TicketTransfer.vue
RiyaMathew-11: Okay - please update the screenshots in PR description too - it shows the typo

RiyaMathew-11:
Okay, I still do feel, it's an add-on workflow when it could be simpler, but we could go with what works best.

PR #425, opened 2024-07-02
Freezing FOSS Profiles

harshtandiya:
Revamp FOSS Profile permissions.

Temporary fix
- Only logging as admin through frappe cloud can enable changing of details now.

image

### Additional Context about Testing

- Used scripts that made use of functions like `get_list` and `set_value` of one user to a different user via CLI. 

- We created test users there with normal logged user perms (without sys manager)

- Then tried running the scripts to test the permission levels of the doctypes. 

- The current permission settings of FOSS User Profiles blocks the effect of these methods. 

Screenshot 2024-07-03 at 1 19 47 PM

### DIFF
--- a/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
+++ b/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
@@ -99,7 +99,8 @@
    "fieldname": "email",
    "fieldtype": "Data",
    "label": "Email",
-   "options": "Email"
+   "options": "Email",
+   "read_only": 1
   },
   {
    "fieldname": "current_city",
@@ -278,7 +279,7 @@
  "index_web_pages_for_search": 1,
  "is_published_field": "is_published",
  "links": [],
- "modified": "2024-07-02 16:17:42.519778",
+ "modified": "2024-07-02 16:21:35.416189",
  "modified_by": "Administrator",
  "module": "FOSS Profiles",
  "name": "FOSS User Profile",
--- a/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
+++ b/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
@@ -279,7 +279,7 @@
  "index_web_pages_for_search": 1,
  "is_published_field": "is_published",
  "links": [],
- "modified": "2024-07-02 16:21:35.416189",
+ "modified": "2024-07-02 17:36:34.436149",
  "modified_by": "Administrator",
  "module": "FOSS Profiles",
  "name": "FOSS User Profile",
@@ -305,14 +305,6 @@
    "read": 1,
    "report": 1,
    "role": "System Manager"
-  },
-  {
-   "read": 1,
-   "role": "Guest"
-  },
-  {
-   "read": 1,
-   "role": "All"
   }
  ],
  "quick_entry": 1,
--- a/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
+++ b/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
@@ -82,7 +82,6 @@
    "fieldtype": "Column Break"
   },
   {
-   "fetch_from": "user.full_name",
    "fieldname": "full_name",
    "fieldtype": "Data",
    "label": "Full Name"
@@ -279,7 +278,7 @@
  "index_web_pages_for_search": 1,
  "is_published_field": "is_published",
  "links": [],
- "modified": "2024-04-06 22:18:47.314419",
+ "modified": "2024-07-02 16:17:42.519778",
  "modified_by": "Administrator",
  "module": "FOSS Profiles",
  "name": "FOSS User Profile",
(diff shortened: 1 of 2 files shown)


fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
```diff
@@ -100,7 +99,8 @@
    "fieldname": "email",
    "fieldtype": "Data",
    "label": "Email",
-   "options": "Email"
+   "options": "Email",
+   "read_only": 1
```
rahulporuri, line 103: so this makes profiles read-only?

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
```diff
@@ -294,21 +294,25 @@
...
    "report": 1,
-   "role": "All",
-   "select": 1,
-   "share": 1,
-   "write": 1
+   "role": "System Manager"
+  },
+  {
+   "read": 1,
+   "role": "Guest"
```
rahulporuri: does this mean guests can read a profile online but not do a get all?

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
```diff
@@ -294,21 +294,25 @@
...
+  },
+  {
+   "read": 1,
+   "role": "All"
```
rahulporuri: does this mean that everyone can read a profile but not do a get all? why the need for two separate "Guest" and "All" with "read"?

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
```diff
@@ -294,21 +294,25 @@
...
    "email": 1,
    "export": 1,
-   "print": 1,
    "read": 1,
    "report": 1,
-   "role": "All",
-   "select": 1,
-   "share": 1,
-   "write": 1
+   "role": "System Manager"
```
rahulporuri, line 307: is System Manager > Administrator or Administrator > System Manager?

rahulporuri:
And just to confirm, does this enable guest to create new profile? because like we discussed, we need that functionality to ensure that people can submit proposals for indiafoss and signup for foss hack

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
mngshm: Administrator > System Manager. 

We can only log in as Administrator through Frappecloud. Administrator is the Super user.

rahulporuri: > We can only log in as Administrator through Frappecloud.

and taking over a foss united profile does not give them access to frappecloud because the authentication for that happens separately. correct?

rahulporuri: Does the system manager need all of these roles? Should they only be able to read and delete?

harshtandiya: Administrator > System Manager

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
harshtandiya: Not this field, but all the permissions listed below do

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
harshtandiya: They can view the details of a profile. The "Select" mostly handles the `get_list` sort of things. Should we disable Read as well for a good measure?

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
harshtandiya: Yes, that is the idea. 

All counts for Registered Users, Guest for Unregistered users.

harshtandiya:
@rahulporuri People can still create profiles, as that is a whole different function on its own

See line 11 : https://github.com/fossunited/fossunited/blob/develop/fossunited/fossunited/user_utils.py

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
rahulporuri: > The "Select" mostly handles the get_list sort of things.

okay. so with "read", an API call can be made to get a specific users' information. right? whether that API call happens via the website or via the CLI is irrelevant. the username is publicly visible on our website in many places.

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
harshtandiya: Gave the permission to email because Ruchika and I often work on ticket corrections and send corrected tickets/invoices through desk, which needs the permission for emails.

export can be removed.

Report is in case of insights, as we are doing a lot of workspaces in desk.

rahulporuri: > Gave the permission to email because Ruchika and I often work on ticket corrections and send corrected tickets/invoices through desk, which needs the permission for emails.

thanks for the explainer. there are User Personas emerging from this discussion that can be used in the future to refine permissions better.

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
harshtandiya: Yes. I enabled read because profiles have web views. Should we disable read for now and find alternative ways to show information on webpage? or alternatively try to find ways to hide critical information better

RiyaMathew-11:
The Read permissions enabled to all - still seems to activate get_list - tried with a custom doctype on the system - and it seems to work

rahulporuri (approved):

RiyaMathew-11:
@rahulporuri @harshtandiya I've updated the PR description with additional context about tests.

PR #426, opened 2024-07-03
chore!: reduce permissions for rsvp, cfp & tickets

harshtandiya:
- Reduce permissions for RSVP doc, RSVP Submissions, CFP Doc and CFP Submissions
- Other than the System user, no person can do a `get_list` on CFP and RSVP submissions. Before this PR, it was easy to get all the RSVP or CFP submissions if the event id was know.
- System User still has permissions for Read/Write/Create/Delete for CFP and RSVP submission as it is heavily used by internal team comprising of system users. Do we want to change it too? I am afraid that may come with it's own set of problems.

- Dashboards will break on this. Working on fixing it in next pr
- Dashboards utilize a lot of get_list, get_value to render the CFP and RSVP submission lists to the organizers. With a change of permissions, the lists will stop getting rendered. One proposed solution: abstract the methods for getting lists such that they have validations that check if the person is an organizer in order to load the list of RSVP/CFP submissions.
- Organizers will not be able to create CFP / RSVP forms temporarily until a fix for this. It will all be done internally for the time being by desk users(our team).

image

---

Also reduced permissions for Tickets, removed the permission for `All` which included `select` `read` and `write`. Now only system manager has the permissions, as per above configuration.

Before, there was a chance that a third party can change the data of any ticket using APIs, that should no longer be the case.

### DIFF
--- a/fossunited/chapters/doctype/foss_event_rsvp/foss_event_rsvp.json
+++ b/fossunited/chapters/doctype/foss_event_rsvp/foss_event_rsvp.json
@@ -126,7 +126,7 @@
  "index_web_pages_for_search": 1,
  "is_published_field": "is_published",
  "links": [],
- "modified": "2024-03-28 15:38:03.136622",
+ "modified": "2024-07-03 10:30:16.158558",
  "modified_by": "Administrator",
  "module": "Chapters",
  "name": "FOSS Event RSVP",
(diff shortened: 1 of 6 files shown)


rahulporuri:
i think it's time to add "security" as one of the options for the issue and PR templates.

RiyaMathew-11:
> Dashboards utilize a lot of get_list, get_value to render the CFP and RSVP submission lists to the organizers.

We could create a new role - FOSS Event Organiser and set them permissions ? Would that work ?

harshtandiya:
@RiyaMathew-11 I am working on the permissions and roles

rahulporuri:
> Other than the System user, no person can do a get_list on CFP and RSVP submissions. Before this PR, it was easy to get all the RSVP or CFP submissions if the event id was know.

We should definitely prevent listall on CFP submissions but what information is revealed from RSVP submissions? Is email of the proposer and other personal details displayed?

harshtandiya:
all the data associated with the doctype are visible, along with owner details. @rahulporuri . Let's get this moving ASAP

NagariaHussain (approved):

PR #429, opened 2024-07-03
Code Cleanup

harshtandiya:
- Removed a lot of redundant code
- Fixed a couple of open whitelisted methods, that could've been dangerous
- Removed `Create a FOSS Profile` web page, as profiles are now auto-created at the time of signup and the page is redundant.


fossunited/api/dashboard.py
```diff
@@ -8,7 +8,7 @@
 
 @frappe.whitelist(allow_guest=True)
 def get_event(name: str) -> dict:
-    return frappe.get_doc("FOSS Chapter Event", name)
+    return frappe.get_doc("FOSS Chapter Event", name).as_dict()
```
NagariaHussain: Why?

fossunited/foss_profiles/doctype/foss_user_profile/templates/foss_user_profile.html
```diff
@@ -380,7 +380,7 @@ <h4 class="mr-1">{{doc.full_name}}</h4>
                 'doctype': table,
             },
             callback: (r) => {
-                d = new frappe.ui.Dialog({
+/*                d = new frappe.ui.Dialog({
```
NagariaHussain: Let's not keep commented our code. That's what source control is for.

fossunited/fossunited/doctype/foss_event_cfp/foss_event_cfp.py
```diff
@@ -199,11 +199,10 @@ def check_if_already_submitted(self):
 @frappe.whitelist()
 def create_cfp_submission(fields):
     fields_dict = {
-        "doctype": "FOSS Event CFP Submission",
         "submitted_by": frappe.session.user,
     }
     fields_dict.update(frappe.parse_json(fields))
-    return create_submission(fields_dict)
+    return create_submission("FOSS Event CFP Submission", fields_dict)
```
NagariaHussain: Why a separate API is required for this? Why no used the out of the box rest API?

fossunited/fossunited/forms.py
```diff
@@ -7,14 +7,32 @@
...
+    """
+    Validate the doctype to be either "FOSS Event CFP Submission" or "FOSS Event RSVP Submission"
+    """
+    if doctype not in [
+        "FOSS Event RSVP Submission",
+        "FOSS Event CFP Submission",
+    ]:
+        return False
+
+    return True
```
NagariaHussain:
```suggestion
    return doctype in (
        "FOSS Event RSVP Submission",
        "FOSS Event CFP Submission",
    )
```

NagariaHussain: Less verbose pattern.

fossunited/fossunited/forms.py
```diff
@@ -7,14 +7,32 @@
...
     Fields variable should have the doctype values as well
     """
-    doc = frappe.get_doc(fields)
+    if not is_valid_doctype(doctype):
+        frappe.log("Unauthorized usage of create submission")
+        frappe.msgprint("Not allowed")
+        return
+
+    doc = frappe.get_doc(doctype, fields)
     doc.insert(ignore_permissions=True)
```
NagariaHussain: Why are we taking the DocType at all as argument here? Let's have separate endpoints if absolutely needed.

fossunited/fossunited/forms.py
```diff
@@ -26,6 +44,11 @@ def update_submission(doctype, submission, fields, custom):
     Used for RSVP and CFPS
     Update the given submission with the given fields and custom answers
     """
+    if not is_valid_doctype(doctype):
+        frappe.log("Unauthorized usage of update_submission")
+        frappe.msgprint("Not allowed")
+        return
```
NagariaHussain: Should frappe.throw with 403.

fossunited/fossunited/forms.py
```diff
@@ -86,6 +109,11 @@ def publish_form(doctype, docname):
     Used to Publish RSVP and CFP forms by the user.
     This bypasses the permissions required to publish the form.
     """
+    if not is_valid_doctype(doctype):
+        frappe.log("Unauthorized usage of publish_form")
+        frappe.throw("Not allowed")
+        return
```
NagariaHussain: Same, throw 403.

fossunited/fossunited/forms.py
```diff
@@ -98,6 +126,11 @@ def unpublish_form(doctype, docname):
     Used to Unpublish RSVP and CFP forms by the user.
     This bypasses the permissions required to unpublish the form.
     """
+    if not is_valid_doctype(doctype):
+        frappe.log("Unauthorized usage of unpublish_form")
+        frappe.throw("Not allowed")
+        return
+
```
NagariaHussain: Lots of repetition of this same set of lines, hinting there may be a better and DRY way.

fossunited/fossunited/utils.py
```diff
@@ -309,78 +101,25 @@ def get_user_editable_doctype_fields(doctype, docname=None):
...
+#         }
+#     )
+#     for field in fields:
+#         doc.set(field, fields[field])
+#     doc.parent = parent
+#     doc.parenttype = parenttype
+#     doc.insert(ignore_permissions=True)
+#     parentdoc = frappe.get_doc(parenttype, parent)
+#     parentdoc.append(parentfield, doc)
+#     parentdoc.save(ignore_permissions=True)
```
NagariaHussain: Again, let's remove commented code.

fossunited/fossunited/utils.py
```diff
@@ -554,5 +274,6 @@ def get_foss_profile(email):
         return None
 
     return (
-        frappe.get_doc("FOSS User Profile", {"user": email}) or None
+        frappe.get_doc("FOSS User Profile", {"user": email}).as_dict()
+        or None
```
NagariaHussain: Why the `or None` part? Get doc will either return something or will throw.

fossunited/www/me.py
```diff
@@ -17,8 +17,8 @@ def get_context(context):
         if redirect_route not in [None, ""]:
             frappe.local.flags.redirect_location = redirect_route
         else:
-            frappe.local.flags.redirect_location = (
-                "/create-foss-profile"
+            frappe.msgprint(
+                "There's been an error in the creation of your FOSS Profile. Please contact administrator"
             )
```
NagariaHussain: Throw?

NagariaHussain (changes requested):

fossunited/api/dashboard.py
harshtandiya: Is it better to return dict rather than object or it doesn't make a difference?

fossunited/foss_profiles/doctype/foss_user_profile/templates/foss_user_profile.html
harshtandiya: Sure, removing it

fossunited/fossunited/doctype/foss_event_cfp/foss_event_cfp.py
harshtandiya: Refactoring

fossunited/api/dashboard.py
NagariaHussain: Does not make a difference, Frappe will do that for you.

fossunited/fossunited/doctype/foss_event_cfp/foss_event_cfp.py
```diff
@@ -198,12 +196,19 @@ def check_if_already_submitted(self):
...
+    if not frappe.db.exists(
+        "FOSS Event CFP", frappe.parse_json(fields).get("linked_cfp")
+    ):
+        frappe.throw("Invalid CFP ID.", frappe.DoesNotExistError)
+
     fields_dict = {
         "doctype": "FOSS Event CFP Submission",
         "submitted_by": frappe.session.user,
     }
     fields_dict.update(frappe.parse_json(fields))
```
NagariaHussain: If I pass in doctype in fields, it will update/ 
Override in the dict  😨

fossunited/fossunited/utils.py
```diff
@@ -553,6 +252,6 @@ def get_foss_profile(email):
     if email in ["guest@example.com", "admin@example.com"]:
         return None
 
-    return (
-        frappe.get_doc("FOSS User Profile", {"user": email}) or None
-    )
+    return frappe.get_doc(
+        "FOSS User Profile", {"user": email}
+    ).as_dict()
```
NagariaHussain: .

NagariaHussain (changes requested):

RiyaMathew-11:
> Removed Create a FOSS Profile web page, as profiles are now auto-created at the time of signup and the page is redundant

At what point does a person create their username, after signup if this page is removed ?

harshtandiya:
@RiyaMathew-11 auto assigned at time of creation 
check: https://github.com/fossunited/fossunited/blob/develop/fossunited/fossunited/user_utils.py

RiyaMathew-11:
I am a bit sure - how feasible this way is. 

Split and join of names can have some pitfalls: 

1. If I take a name, Riya Mathew - and someone else takes the same name - that would create duplicate usernames? 

2. A question - then which profile page would the route redirect to?

harshtandiya:
@RiyaMathew-11 that is already handled, see the function above it

NagariaHussain (approved):

PR #440, opened 2024-07-07
chore(ticket-transfer): better api endpoints & permissions

harshtandiya:
- [x] ⚙️Chore
- [x] 🧑‍💻Refactor

## Description

- Fixed an api endpoint to get profile data of the session user for dashboard header
- Refactored API endpoints to work with the new permissions for Ticket Transfer doctype.
- Less usage of `frappe.client.get` as it won't work with the new permissions. 
- New API endpoints make use of `frappe.db.get_value` to get only the required value, nothing more.
- Permissions given to Guest and All as follows:
This is done because write permissions are required to run the method when status is changed to `Completed`. `frappe.db.set_value` method won't call ORM triggers like `validate` and `before_save`.
image
- Validations added for transfer doctype which run at the time of creation. They check whether the transfer being created has a valid ticket.
- The validation also happens to ensure that the status at the time of creation is `Pending Approval` and nothing else.

### DIFF
--- a/fossunited/api/tickets.py
+++ b/fossunited/api/tickets.py
@@ -11,11 +11,8 @@ def check_ticket_validity(ticket_id: str):
     Check if the ticket is valid or not
     """
     is_ticket_valid = frappe.db.exists("FOSS Event Ticket", ticket_id)
-    print(is_ticket_valid)
-    if is_ticket_valid:
-        return True
 
-    return False
+    return bool(is_ticket_valid)
 
 
 @frappe.whitelist(allow_guest=True)
@@ -56,21 +53,19 @@ def create_transfer_request(ticket: str, receiver_details: dict):
 
 
 @frappe.whitelist()
-def check_transfer_doc_validity(transfer_id: str):
+def get_transfer_doc_validity(transfer_id: str):
     """
     Check the validity of transfer doc/id
     """
     is_valid_id = frappe.db.exists(
         "FOSS Event Ticket Transfer", transfer_id
     )
-    if is_valid_id:
-        return True
 
-    return False
+    return bool(is_valid_id)
 
 
 @frappe.whitelist()
-def get_transfer_doc(id: str):
+def get_transfer_details(id: str):
     """
     Get the transfer doc
     """
--- a/fossunited/api/tickets.py
+++ b/fossunited/api/tickets.py
@@ -1,0 +1,94 @@
+"""
+APIs for Tickets and Transfer Tickets
+"""
+
+import frappe
+
+
+@frappe.whitelist(allow_guest=True)
+def check_ticket_validity(ticket_id: str):
+    """
+    Check if the ticket is valid or not
+    """
+    is_ticket_valid = frappe.db.exists("FOSS Event Ticket", ticket_id)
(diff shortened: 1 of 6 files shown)


fossunited/ticketing/doctype/foss_event_ticket_transfer/foss_event_ticket_transfer.py
```diff
@@ -30,22 +30,44 @@ class FOSSEventTicketTransfer(Document):
     # end: auto-generated types
     pass
 
+    def before_insert(self):
+        self.validate_ticket_exists()
+        self.check_status()
```
NagariaHussain:
```suggestion
        self.validate_status_is_pending()
```

fossunited/api/tickets.py
```diff
@@ -0,0 +1,94 @@
...
+
+@frappe.whitelist()
+def change_transfer_status(transfer_id: str, status: str):
+    """
+    Change the status of the transfer request
+    """
+    doc = frappe.get_doc("FOSS Event Ticket Transfer", transfer_id)
+    doc.status = status
+    doc.save()
+    return True
```
NagariaHussain, line 89: How is any different than db.set_value?

fossunited/api/tickets.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    Get the transfer doc
+    """
+    doc = frappe.db.get_value(
+        "FOSS Event Ticket Transfer",
+        id,
+        ["name", "status", "ticket"],
+        as_dict=True,
+    )
+    return doc
```
NagariaHussain: The name of the function is misleading. This function only returns some values not the doc object.

fossunited/api/tickets.py
```diff
@@ -0,0 +1,94 @@
...
+    Check the validity of transfer doc/id
+    """
+    is_valid_id = frappe.db.exists(
+        "FOSS Event Ticket Transfer", transfer_id
+    )
+    if is_valid_id:
+        return True
+
+    return False
+
```
NagariaHussain:
```suggestion
   return bool(is_valid_id)
```

Frequently used pattern, good to keep in mind.

fossunited/api/tickets.py
```diff
@@ -0,0 +1,94 @@
...
+            "tshirt_size": receiver_details.get("tshirt_size"),
+            "status": "Pending Approval",
+        }
+    )
+    transfer_request.insert(ignore_permissions=True)
+    return transfer_request
+
+
+@frappe.whitelist()
+def check_transfer_doc_validity(transfer_id: str):
```
NagariaHussain: Again, the function name should clarify the intent/what's it doing.

check/validate implies it will be some kind of validate and throw function. This one just returns whether the provided id is valid. 

And according to the document I shared, this should start with `get_`

NagariaHussain (changes requested):

fossunited/api/tickets.py
harshtandiya: There are controller methods associated with Tranfer Doctype which won't run if I do a set_value

NagariaHussain (approved):

PR #447, opened 2024-07-09
[TEST] add unit tests for ticket-transfer doctype

harshtandiya:
- [x] Test
## Description
This PR adds a couple of unit tests for the Ticket Transfer Doctype.
- Testing Ticket Transfer Workflow ->
   - Check that ticket is created only for a valid event.
   - Check that once the transfer is done, there is no ticket remaining with the details of the previous owner
   - Verify that the ticket has the details of the new owner after transfer 
- Testing Transfer Status to be "Pending Approval" at the time of creation, and that it throws a ValidationError if it is anything else.

```
----------------------------------------------------------------------
Ran 2 tests in 1.889s

```

### DIFF
--- a/fossunited/ticketing/doctype/foss_event_ticket_transfer/test_foss_event_ticket_transfer.py
+++ b/fossunited/ticketing/doctype/foss_event_ticket_transfer/test_foss_event_ticket_transfer.py
@@ -50,7 +50,6 @@ class TestFOSSEventTicketTransfer(FrappeTestCase):
         # Change status of transfer to completed
         transfer.status = "Completed"
         transfer.save()
-        self.assertEqual(transfer.status, "Completed")
 
         # Then check that a ticket with older credentials does not exists.
         # Also check that ticket has new details.
@@ -73,10 +72,6 @@ class TestFOSSEventTicketTransfer(FrappeTestCase):
             },
         )
         self.assertTrue(new_ticket_exists)
-
-        transfer.delete()
-        ticket.delete()
-        event.delete()
 
     def test_check_status_pending_on_create(self):
         # Given an event and a ticket linked to the event
--- a/fossunited/ticketing/doctype/foss_event_ticket_transfer/test_foss_event_ticket_transfer.py
+++ b/fossunited/ticketing/doctype/foss_event_ticket_transfer/test_foss_event_ticket_transfer.py
@@ -1,8 +1,79 @@
 # Copyright (c) 2024, Frappe x FOSSUnited and Contributors
 # See license.txt
 
+from datetime import datetime, timedelta
+
+import frappe
 from frappe.tests.utils import FrappeTestCase
 
 
 class TestFOSSEventTicketTransfer(FrappeTestCase):
-    pass
+    def test_ticket_transfer(self):
+        # Given for an event, a ticket is created. For that ticket, a transfer is generated.
+        event = frappe.get_doc(
+            {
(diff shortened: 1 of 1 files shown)


fossunited/ticketing/doctype/foss_event_ticket_transfer/test_foss_event_ticket_transfer.py
```diff
@@ -1,8 +1,119 @@
...
+        )
+        transfer.insert()
+
+        # When Ticket Transfer is created, status should be pending
+        self.assertEqual(transfer.status, "Pending Approval")
+
+        # Change status of transfer to completed
+        transfer.status = "Completed"
+        transfer.save()
+        self.assertEqual(transfer.status, "Completed")
```
NagariaHussain: You don't have to test the framework 😅

fossunited/ticketing/doctype/foss_event_ticket_transfer/test_foss_event_ticket_transfer.py
```diff
@@ -1,8 +1,119 @@
...
+                "email": "rahul@test.xyz",
+                "full_name": "Rahul",
+                "event": event.name,
+            },
+        )
+        self.assertTrue(new_ticket_exists)
+
+        transfer.delete()
+        ticket.delete()
+        event.delete()
```
NagariaHussain: Rollback automatically happens...

NagariaHussain (dismissed):

fossunited/ticketing/doctype/foss_event_ticket_transfer/test_foss_event_ticket_transfer.py
rahulporuri: after each test or after the entire testsuite?

we wanted to ensure that we don't have potential test interactions between unit tests, which is why we went ahead and deleted them manually

fossunited/ticketing/doctype/foss_event_ticket_transfer/test_foss_event_ticket_transfer.py
rahulporuri: this might be a left over assert - at first, we were using a method on the doctype to set the status and perform the validation internally. we ran into some issues there so this assert can go now.

fossunited/ticketing/doctype/foss_event_ticket_transfer/test_foss_event_ticket_transfer.py
```diff
@@ -1,8 +1,119 @@
...
+
+    def test_check_status_pending_on_create(self):
```
rahulporuri: nitpicky comment - we might need a better name because it wasn't obvious what we were testing until i read the test code. but i cant come up with a better name myself so 🤷

harshtandiya: `test_status_pending_on_create`?

rahulporuri (approved):
LGTM. I think it's worth waiting for @NagariaHussain s review for a day or two. We're not in a hurry to merge this PR.

rahulporuri:
@NagariaHussain we're going ahead and merging this PR for now. Please do take a look at it post-merge, if you have the time. we are still trying our best to understand the best practices around unittesting so your comments in this regard are incredibly valuable

NagariaHussain (approved):

PR #453, opened 2024-07-09
fix: bug-fixes in hackathon module

### LINKED ISSUE #459: Missing Values in Requests Fetch in Hackathon Organisers' Dashboard

RiyaMathew-11:
- [x] ⚙️Chore
- [x] 🐛Bug Fix

## Description

- [x] Fixed Nonetype error for fresh localhost addition
- [x] String - to lower error while setting hackathon routes
- [x] Fixed details-not-fetched error in localhost organiser dashboard
    - The `get_project_by_team` method created a breakpoint for teams that did not have projects created, as it did not handle a None state.
    - This caused a `DoesNotExistError`, breaking further actions.
    - Added a `try-catch` block to handle this condition gracefully when the method is called in `get_localhost_requests_by_team` method. 
    - This ensures that if a team does not have a project, the code can handle the None state appropriately without breaking.
      
- [x] Updated the CTA in Localhosts requests list from `View Project` to `Project Title`
- [x] Added a small helper function that truncates strings in cases needed.

## Related Issues & Docs

- Closes #459

## Checklist
- [x] ~The code follows the project's coding standards.~

local-host-org-dashboard

truncate-case


RiyaMathew-11:
Screenshot 2024-07-15 at 6 36 48 PM


It seems a bigger project_title would break this style too, fixing it here too @harshtandiya. 

Instead of truncate, better would clipping to newline

harshtandiya:
@RiyaMathew-11 rather than truncate, please wrap it so it comes down

RiyaMathew-11:
> @RiyaMathew-11 rather than truncate, please wrap it so it comes down

Yeah - i was editing the comment - just did

RiyaMathew-11:
Screenshot 2024-07-15 at 6 58 34 PM

Updated.

Tho @harshtandiya - is uppercase style needed there - it looks a bit off I think ?

harshtandiya:
@RiyaMathew-11 it's fine

dashboard/src/components/localhost/AttendeeRequestList.vue
```diff
@@ -162,14 +162,32 @@
...
           </div>
-          <div v-else-if="column.label == 'Project'">
-            <Button
-              size="sm"
-              variant="ghost"
-              @click="redirectToProfile(row.project_route)"
+          <div class="flex" v-else-if="column.label == 'Project'">
+            <a
+              v-if="row.project_route"
+              :href="row.project_route"
```
harshtandiya: This won't work. It will redirect to the route such as `dashboard/<project_route>`

dashboard/src/components/localhost/AttendeeRequestList.vue
```diff
@@ -195,6 +213,7 @@ import {
 } from 'frappe-ui'
 import { ref } from 'vue'
 import RequestDetailDialog from '@/components/localhost/RequestDetailDialog.vue'
+import { truncateStr } from '../helpers/utils'
```
harshtandiya: `@/components/helpers/utils` might be better?

harshtandiya (changes requested):

dashboard/src/components/helpers/utils.js
```diff
@@ -0,0 +1,4 @@
+export const truncateStr = (title, len) => {
+
+    return title.length > len ? title.substring(0, len) + "..." : title
+    }
```
rahulporuri: nitpicky comment - is this file formatted using prettier? the formatting feels a bit wonky to my eye 😅

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/foss_hackathon_localhost.py
```diff
@@ -49,7 +49,9 @@ def assign_localhost_organizer_role(self):
 
     def check_if_member_removed(self):
         prev_doc = self.get_doc_before_save()
-        if not len(prev_doc.organizers) > len(self.organizers):
+        if not prev_doc or not len(prev_doc.organizers) > len(
+            self.organizers
+        ):
```
rahulporuri: I would be careful with operator precedence here. Is the above code what we mean to do here? I don't remember off the top of my head but is this statement `if not (x or (not y) > z )` or `if not (x or (not (y > z) )` if (not x) or (not (y > z))`?

i think you mean to use the last one i.e. `if (not x) or (not (y > z))`. is that correct?

if so, could we simplify this using boolean algebra? again, i dont remember it off the top of my head but isn't `(not a) or (not b) == not (a and b)`? the latter reads better to me, without the double negatives.

what do you think @RiyaMathew-11 ?

dashboard/src/components/localhost/AttendeeRequestList.vue
RiyaMathew-11: The helper dir shouldn't be inside components - I have updated the path

dashboard/src/components/localhost/AttendeeRequestList.vue
RiyaMathew-11: Made the changes - also made it a helper function in utils - since it was written redundantly in all files where it's called.

dashboard/src/components/helpers/utils.js
RiyaMathew-11: Oohh no I missed this - did now.

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/foss_hackathon_localhost.py
RiyaMathew-11: Yep, it's the third one - `if (not x) or (not (y > z))`. 

And yes, simplifying it makes it readable and more clear too, changing it:

```
        if not (prev_doc and len(prev_doc.organizers) > len(
            self.organizers)
        ):
```

harshtandiya (approved):

PR #457, opened 2024-07-11
chore: add permissions for `CFP Reviewer` Role

harshtandiya:
- [x] ⚙️Chore

## Description

- Introduce a new role `CFP Reviewer`
- Users assigned `CFP Reviewer` roles will be having the write permissions to level 0 and level 1 fields of CFP Submission doctype. This includes level 1 field `reviews` child table which has all the reviews for the cfp.
- Methods introduced which assign `CFP Reviewer` Role when a user is added to the Global CFP Reviewer List

- Added unit tests to FOSS Global CFP Review Settings doctype. These tests: 1. Creation of the single doctype without any members. 2. Role assignment on addition of a member. 3. Role de-assignment on removal of the member

### DIFF
--- a/fossunited/fossunited/doctype/foss_event_cfp/foss_event_cfp.py
+++ b/fossunited/fossunited/doctype/foss_event_cfp/foss_event_cfp.py
@@ -43,8 +43,6 @@ class FOSSEventCFP(WebsiteGenerator):
         reviewers = frappe.get_single(
             "FOSS Global CFP Review Settings"
         ).members
-        print(reviewers)
-
         for reviewer in reviewers:
             self.append(
                 "cfp_reviewers",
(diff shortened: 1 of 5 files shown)


fossunited/fossunited/doctype/foss_event_cfp/foss_event_cfp.py
```diff
@@ -43,8 +43,6 @@ def assign_reviewers(self):
         reviewers = frappe.get_single(
             "FOSS Global CFP Review Settings"
         ).members
-        print(reviewers)
```
rahulporuri, line 46: there is another `print` call in the `foss_chapter_event` module that can be removed

harshtandiya: This was removed. Will be removing that too

harshtandiya:
@rahulporuri removed print statements

rahulporuri (approved):

PR #494, opened 2024-07-22
feat(rsvp): auto close rsvp on max submission count

### LINKED ISSUE #482: RSVP Form doesn't unpublish if max RSVP count exceeded
### Describe the bug

The FOSS United Platform has a field to set the maximum RSVP count but its not taking the maximum RSVP count into consideration and allows RSVPs even after the count is exceeded

image

However still 116 people have registered 👇🏻 
Screenshot 2024-07-19 at 11 37 25 AM

### Expected behavior

The RSVP form should unpublish or set a message that "RSVP exceeded the maximum capacity" and strictly disallow RSVP post exceeded.

### To Reproduce

Set Max RSVP count and try to register after the count is exceeded

harshtandiya:
- [x] 🍕Feature

## Description

- Until now, the `max_rsvp_count` was present just for statistics, and was not doing much. 
- Added a method in `after_save` controller of RSVP Submission doctype, such that, after every submission, the max count is checked with the count of rsvp submissions for that particular event. If `submission_count >= max_rsvp_count` the RSVP form is unpublished.
- ~Also added a function that validates if the rsvp form is published. If not, the user is not allowed to create a submission and an error gets thrown.~
- Added test for testing form unpublish when max submission count is hit.

## Related Issues & Docs

closes #482

### DIFF
--- a/fossunited/chapters/doctype/foss_event_rsvp_submission/foss_event_rsvp_submission.py
+++ b/fossunited/chapters/doctype/foss_event_rsvp_submission/foss_event_rsvp_submission.py
@@ -33,9 +33,6 @@ class FOSSEventRSVPSubmission(Document):
     # end: auto-generated types
     pass
 
-    def before_insert(self):
-        self.validate_form_published()
-
     def validate(self):
         self.validate_linked_rsvp_exists()
 
@@ -62,16 +59,6 @@ class FOSSEventRSVPSubmission(Document):
             "FOSS Event RSVP", self.linked_rsvp, "max_rsvp_count"
         )
         return max_count
-
-    def validate_form_published(self):
-        is_published = frappe.db.get_value(
-            "FOSS Event RSVP", self.linked_rsvp, "is_published"
-        )
-
-        if not is_published:
-            frappe.throw(
-                "RSVP form is closed", frappe.ValidationError
-            )
 
     def validate_linked_rsvp_exists(self):
         if not frappe.db.exists("FOSS Event RSVP", self.linked_rsvp):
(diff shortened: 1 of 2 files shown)


RiyaMathew-11:
A meta comment/question: 

If an RSVP form is unpublished, it does get reflected in the UI right? If yes, then what are the chances for the user to attempt to RSVP again? 

If a form is unpublished and the user is trying to access it, doesn't the system consider the link as invalid and reverts to 404. 

If both these states hold, why is the unpublished rsvp validation required on the user side?

harshtandiya:
- The page throws a 404 when the form is unpublished.
- The event pages show when the form has been unpublished.
- Unpublished validation are required so that no one does a `client.insert`

RiyaMathew-11:
> so that no one does a client.insert

Can that attempt be made still after we have set permissions for chapter team member and system manager ? 

Screenshot 2024-07-22 at 4 35 10 PM

harshtandiya:
I guess its redundant then

fossunited/chapters/doctype/foss_event_rsvp_submission/foss_event_rsvp_submission.py
```diff
@@ -33,9 +33,46 @@ class FOSSEventRSVPSubmission(Document):
...
+                False,
+            )
+
+    def get_max_count(self):
+        max_count = frappe.db.get_value(
+            "FOSS Event RSVP", self.linked_rsvp, "max_rsvp_count"
+        )
+        return max_count
+
+    def validate_form_published(self):
```
RiyaMathew-11: This method might not be needed. 

Since this is a bit redundant as there are other methods that reflect an unpublish with a UI change and permission levels in the submission doctype that blocks a direct insert by client API.

RiyaMathew-11:
> I guess its redundant then

Okay! @harshtandiya I've added my comment based on this.

RiyaMathew-11 (approved):

Issue #364, opened 2024-06-04
[FEATURE] better error messages


mngshm:
### Feature description

At the moment, on the dashboard there is only an error with the default error message "Failed to update details" or something like that while creating or managing events. I think it would be better if we define where the error messagef exactly is or if there is some empty field.

rahulporuri:
closing as partially addressed

PR #523, opened 2024-08-06
chore(event): Add validation for event permalink

### LINKED ISSUE #509: More rigid permalink rules
### Issue description

using `/` in permalinks of events seem to break the cfp and rsvp all pages. Create validations for the format of permalinks

### Steps to reproduce (if applicable)

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

harshtandiya:
- [x] ⚙️Chore

## Description

- Added a validation which does not let people create events with a permalink having the `/` character

## Related Issues & Docs

closes #509

### DIFF
--- a/fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
+++ b/fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
@@ -134,6 +134,12 @@ class FOSSChapterEvent(WebsiteGenerator):
                 "Event Permalink cannot contain / character. Please use a different permalink."
             )
 
+        # ensure permalink is only alphanumeric with exception of -
+        if not self.event_permalink.replace("-", "").isalnum():
+            frappe.throw(
+                "Event Permalink can only contain alphabets, numbers and hyphens."
+            )
+
     def update_published_status(self):
         if self.status == "Draft" or self.status == "Cancelled":
             self.is_published = 0
--- a/fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
+++ b/fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
@@ -111,6 +111,25 @@ class FOSSChapterEvent(WebsiteGenerator):
             self.update_published_status()
         self.set_route()
 
+    def validate(self):
+        self.validate_permalink()
+
+    def validate_permalink(self):
+        if self.is_external_event:
+            return
+
+        if frappe.db.exists(
+            self.doctype, {"event_permalink": self.event_permalink}
+        ):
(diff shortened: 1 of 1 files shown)


fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
```diff
@@ -111,6 +111,25 @@ def before_save(self):
...
+
+        if frappe.db.exists(
+            self.doctype, {"event_permalink": self.event_permalink}
+        ):
+            frappe.throw(
+                f"Event Permalink {self.event_permalink} already exists!"
+            )
+
+        if "/" in self.event_permalink:
+            frappe.throw(
```
RiyaMathew-11, line 133: With the validation of `/` - should we try for a more stricter validation pattern ? 
If someone can add a permalink like - `f0$$h@(k` - should we make it with alphanumeric chars only ? 
Screenshot 2024-08-08 at 12 49 19 PM


And maybe with a length cap as well too, so the permalinks don't go too long?

RiyaMathew-11 (dismissed):
Found one bug - need to fix that up.

fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
```diff
@@ -111,6 +111,25 @@ def before_save(self):
...
+        self.validate_permalink()
+
+    def validate_permalink(self):
+        if self.is_external_event:
+            return
+
+        if frappe.db.exists(
+            self.doctype, {"event_permalink": self.event_permalink}
+        ):
+            frappe.throw(
```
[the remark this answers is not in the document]
RiyaMathew-11: This needs to be fixed with an additional check to avoid the same document's permalink - while checking.

RiyaMathew-11, line 128: This sort of takes into account for every save - when I save a permalink, and then makes another change, and then save - the permalink check raises an `Exists Error` 
permalink-error

RiyaMathew-11 (changes requested):
Found one bug - need to fix that.

RiyaMathew-11:
@harshtandiya perhaps a better error message for the dashboard would be nice too, for the change in permalink errors that may come: 

Screenshot 2024-08-08 at 1 35 54 PM

harshtandiya:
> @harshtandiya perhaps a better error message for the dashboard would be nice too, for the change in permalink errors that may come:
> 
> Screenshot 2024-08-08 at 1 35 54 PM

will be handled in #364

RiyaMathew-11 (approved):

PR #525, opened 2024-08-06
fix(ticket-transfer): emails not being sent automatically on transfer

harshtandiya:
## What type of PR is this? (check all applicable)
- [x] 🐛Bug Fix
- [x] 🧑‍💻Refactor

## Description

- Replaced the database methods with documents methods. Reason: database methods don't run controller methods attached to the doctype. Hence, the notifications set up on the desk were also not being sent when the value of `is_transfer_ticket` checkbox was changing for Ticket DocType.
- To bypass permissions set on ticket doctype, the user is temporarily switched to `Administrator` to make changes to ticket and save it. This is achieved using `frappe.set_user`.

## Related Issues & Docs

closes #521

Issue #334, opened 2024-05-15
[BUG] Essential fixes list on the community events dashboard


RiyaMathew-11:
### Describe the bug

Referenced from this discussion: #324 

- [x] Make event publishing workflow connected to status changes

   - Proposed status labels: Draft, Live, Concluded, Cancelled
   - Event will be published if the status is live/concluded

- [x] Standard routing guideline for event links

- [x] Fix typos and mark mandatory fields needed.

- [x] A lead can add or remove members, but do not have the provision to edit roles. 

![Screenshot 2024-05-16 at 1 23 45 AM](https://github.com/fossunited/fossunited/assets/60656677/be79bed6-40e9-41ae-af48-dd46edffedfe)

- [x] The drop down for select members to add - potentially gives a list of all website users  - this isnt a good practice, explore to see if this can be a data field and not a drop down

   - The username/email can be entered - if the user exists, get's added else gives an error alert - `User does not exist`

- [x] Cross-verify removal permissions - the volunteer seem to be able to remove lead (PS: could be because of my desk access)

- [x] The event description field needs to be bigger - to give detailed descriptions if needed - tried one for Kochi meetup: 

   - The `<br/>` tags weren't working on the event page
   - Would be good if we can make the field as the rich text editor

![Screenshot 2024-05-16 at 1 44 48 AM](https://github.com/fossunited/fossunited/assets/60656677/dec33d13-3b31-4107-a7fe-2e495f111f3b)

![Screenshot 2024-05-16 at 1 52 08 AM](https://github.com/fossunited/fossunited/assets/60656677/3dbb1fbb-9f37-4e49-b82c-80048528da72)

(PS: Made it the needed way from desk later)

Edit: Issue referenced in #346 

- [x] The session of event dashboard logs out too fast - it got signed out within 10 min of signin and then had to login again.

harshtandiya:
>  The drop down for select members to add - potentially gives a list of all website users - this isnt a good practice, explore to see if this can be a data field and not a drop down
 The username/email can be entered - if the user exists, get's added else gives an error alert - User does not exist


We are not changing to this, as I don't see a problem with the current thing? There's anyway going to be a public page showing all the foss profiles in future, and hopefully a way to "Follow" or something of similar lines.


> Cross-verify removal permissions - the volunteer seem to be able to remove lead (PS: could be because of my desk access)

A team member can remove anyone other than himself. Moreover, this feature may be too much unwanted complexity. Not changing this atm.

[...]

PR #535, opened 2024-08-08
fix(dashboard): broken chapter member & event volunteer addition workflow

harshtandiya:
- [x] 🐛Bug Fix
- [x] 🧑‍💻Refactor

## Description

- Member addition and volunteer addition workflow were broken for organizer dashboard due to the changes in permissions.
- This PR fixes the workflow to be usable again.
- Added a new vue component `AddMember.vue` which is essentially a dialog box used for adding members for chapter and event both.
- Refactored document controller methods for chapters, which were used to assign and de-assign `Chapter Team Member` and `Chapter Lead` roles to members of the chapter. Refactored the code to use `set_user`, so that the user is switched with Administrator at the time of any save to the chapter document. This is necessary, because Role addition works directly on the `User` doctype which can only be changed by the Administrator.
- Also, added a validation on the dashboard which checks whether the session user is the member of the chapter whose dashboard page is being accessed. If not, it throws an error dialog and redirects the user to dashboard.

### DIFF
--- a/dashboard/src/components/chapter/AddMember.vue
+++ b/dashboard/src/components/chapter/AddMember.vue
@@ -1,0 +1,82 @@
+<template>
+  <Dialog
+    :options="{
+      title: 'Add New Member',
+    }"
+  >
+    <template #body-content>
+      <div class="flex flex-col gap-2">
+        <div class="text-p-base text-gray-700">
+          Enter the username of the new member you want to add to the team.
+        </div>
+        <Autocomplete
+          :options="memberOptions.data"
+          v-model="newMembers"
+          placeholder="Search for a user"
+          :multiple="true"
+        >
+          <template #item-prefix="{ option }">
+            <Avatar
+              shape="circle"
+              :image="option.avatar"
+              :label="option.label"
+              size="lg"
+            />
+          </template>
+        </Autocomplete>
+      </div>
+    </template>
+    <template #actions>
+      <div class="grid grid-cols-2 gap-3">
+        <Button label="Cancel" @click="$emit('close-dialog')" />
+        <Button
+          label="Add"
+          variant="solid"
+          @click="$emit('update:add-member', newMembers)"
+        />
+      </div>
+    </template>
+  </Dialog>
+</template>
+<script setup>
+import { Dialog, Autocomplete, createResource, Avatar } from 'frappe-ui'
+import { ref, defineProps, defineEmits } from 'vue'
+
+const props = defineProps({
+  chapter: Object,
+})
+
+const emits = defineEmits(['update:add-member', 'close-dialog'])
+
+const memberOptions = createResource({
+  url: 'fossunited.api.dashboard.get_user_profile_list',
+  makeParams() {
+    return {
+      filters: {
+        name: [
+          'not in',
+          props.chapter.doc.chapter_members
+            .map((member) => member.chapter_member)
+            .join(','),
+        ],
+      },
+    }
+  },
+  auto: true,
+  realtime: true,
+  transform(data) {
+    return data.map((user) => {
+      return {
+        value: user.name,
+        label: user.username,
+        description: user.full_name,
+        avatar: user.profile_photo
+          ? user.profile_photo
+          : '/assets/fossunited/images/defaults/user_profile_image.png',
+      }
+    })
+  },
+})
+
+const newMembers = ref([])
+</script>
--- a/dashboard/src/components/chapter/AddMember.vue
+++ b/dashboard/src/components/chapter/AddMember.vue
@@ -40,11 +40,24 @@
 </template>
 <script setup>
 import { Dialog, Autocomplete, createResource, Avatar } from 'frappe-ui'
-import { ref, defineProps, defineEmits } from 'vue'
+import { ref, defineProps, defineEmits, computed } from 'vue'
 
 const props = defineProps({
   chapter: Object,
+  event: Object,
 })
+
+const existingMembers = computed(() => {
+  if (props.chapter) {
+    return props.chapter.doc.chapter_members.map((member) => member.chapter_member).join(',')
+  }
+  else if (props.event) {
+    return props.event.doc.event_members.map((member) => member.member).join(',')
+  }
+
+  return []
+})
+
 
 const emits = defineEmits(['update:add-member', 'close-dialog'])
 
@@ -55,9 +68,7 @@ const memberOptions = createResource({
       filters: {
         name: [
           'not in',
-          props.chapter.doc.chapter_members
-            .map((member) => member.chapter_member)
-            .join(','),
+          existingMembers.value
         ],
       },
     }
--- a/dashboard/src/pages/ChapterMembers.vue
+++ b/dashboard/src/pages/ChapterMembers.vue
@@ -1,5 +1,9 @@
 <template>
-  <div v-if="chapter.doc" class="px-4 py-8 md:p-8 w-full z-0 min-h-screen">
+  <div
+    v-if="chapter.doc"
+    class="px-4 py-8 md:p-8 w-full z-0 min-h-screen"
+    :class="chapter.get.loading ? 'animate-pulse' : ''"
+  >
     <div class="flex justify-between mt-4">
       <ChapterHeader :chapter="chapter" />
     </div>
@@ -20,7 +24,10 @@
         v-for="member in chapter.doc.chapter_members"
         :title="member.full_name"
       >
-        <template v-if="member.email != session.user" #actions>
+        <template
+          v-if="member.email != session.user && member.role != 'Lead'"
+          #actions
+        >
           <Button
             theme="red"
             label="Remove"
@@ -40,8 +47,11 @@
         </div>
       </Card>
     </div>
+
+    <!-- REMOVE A MEMBER -->
     <div v-if="selectedMember">
       <Dialog
+        class="z-50"
         v-model="showRemoveModal"
         :options="{
           title: 'Remove Team Member?',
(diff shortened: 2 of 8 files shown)


fossunited/chapters/doctype/foss_chapter/foss_chapter.py
```diff
@@ -57,13 +57,73 @@ def before_save(self):
...
+            frappe.session.data = current_session_data
+
+    def handle_member_removal(self):
+        prev_doc = self.get_doc_before_save()
+        if not prev_doc:
+            return
+
+        current_user = frappe.session.user
+        current_session_data = frappe.session.data
+        frappe.set_user("Administrator")
```
RiyaMathew-11: Tangential doubt: Can this be applied here too ? 

https://github.com/fossunited/fossunited/pull/525/files/8d9bfdfdc9f87e06571060bc5c610816101295c0#r1705559152

@NagariaHussain @harshtandiya

harshtandiya: I don't think `add_roles` & `remove_role` work with `ignore_permission`

dashboard/src/components/chapter/AddMember.vue
```diff
@@ -0,0 +1,93 @@
+<template>
+  <Dialog
+    :options="{
+      title: 'Add New Member',
+    }"
+  >
+    <template #body-content>
+      <div class="flex flex-col gap-2">
+        <div class="text-p-base text-gray-700">
+          Enter the username of the new member you want to add to the team.
```
RiyaMathew-11, line 10: Personal thought: I am not in favour of a considerable user list being exposed here. 
If we could do without the dropdown - would be nice. 

The list could potential also have folks who have anonymised their profiles ?
Thoughts on this @rahulporuri ?

Screenshot 2024-08-10 at 12 36 38 AM

fossunited/chapters/doctype/foss_chapter/foss_chapter.py
RiyaMathew-11: I also have a doubt whether we need to give the Admin permissions here. 

The role addition is anyway executed in the Administrator level right? If it's server side code. We did the same for localhost role additions ? Afaik we did not make the user switch there, and it worked, when a user was added to the organiser list, it triggered the execution of role addition at system level?

fossunited/api/chapter.py
```diff
@@ -0,0 +1,38 @@
+import frappe
+
+from fossunited.doctype_ids import USER_PROFILE
+
+
+@frappe.whitelist()
+def check_if_chapter_member(
```
RiyaMathew-11, line 7: We are checking with this on the frontend - where access to visibly see another chapter dashboard is blocked. 

How about the backend side? With read and write permissions enabled for a `Chapter Team Member`, with the use of APIs: 

1. Can a member of chapter member of A with roles assigned, assign himself as member of chapter B.

2. By this also gain access to events of chapter B ?

RiyaMathew-11: Would a check `before every save` in doctype help to validate if the edit is made by the member of the chapter ? This might potentially block the access via API methods ?

RiyaMathew-11: Screenshot 2024-08-10 at 1 43 52 AM
The permissions set for `All` needs to be changed too @harshtandiya ?

dashboard/src/pages/EventVolunteers.vue
```diff
@@ -60,52 +65,29 @@
       >
       </Dialog>
     </div>
+
     <div v-if="showAddmodal">
-      <Dialog
+      <AddMemberDialog
```
RiyaMathew-11, line 70: Same thoughts here for the display of all users in the drop-down.

RiyaMathew-11 (commented):
Some doubts and thought questions added.

dashboard/src/pages/ChapterMembers.vue
```diff
@@ -20,7 +24,10 @@
         v-for="member in chapter.doc.chapter_members"
         :title="member.full_name"
       >
-        <template v-if="member.email != session.user" #actions>
+        <template
+          v-if="member.email != session.user && member.role != 'Lead'"
+          #actions
+        >
```
RiyaMathew-11, line 30: Screenshot 2024-08-10 at 2 12 34 AM

Any member other than Lead shouldn't have the option to remove another member. They however can have the permission to leave one's own.

RiyaMathew-11 (changes requested):
It's not a nice practice to provide remove member permissions to anyone other than Chapter Lead.

fossunited/chapters/doctype/foss_chapter/foss_chapter.py
harshtandiya: No, the code is not executed on administrator level on server side. I tried it, it will throw a Permission error on User doctype.

harshtandiya: For the LocalHost, the System managers were the one who added the LocalHost members. System managers can make changes to the User doctype

dashboard/src/components/chapter/AddMember.vue
harshtandiya: People with anonymous profile still need to be added to the organizer list, right? Should they choose to organize things. 

Moreover, I don't think this is a bad thing, given that there was a plan to create a central page of all Foss profiles. 

Also, is a profile is "Private" people cannot view it even if they visit the profile route.

fossunited/api/chapter.py
harshtandiya: All shouldn't be a problem, as the chapter's info is public anyway

harshtandiya: > We are checking with this on the frontend - where access to visibly see another chapter dashboard is blocked. 
> 
> How about the backend side? With read and write permissions enabled for a `Chapter Team Member`, with the use of APIs: 
> 
> 1. Can a member of chapter member of A with roles assigned, assign himself as member of chapter B.
> 
> 2. By this also gain access to events of chapter B ?

Can we reference this in a new issue? This might bring additional errors. Won't be taking this up in this PR

dashboard/src/pages/EventVolunteers.vue
harshtandiya: It is fine in my opinion. The name and username is all that is shown.

dashboard/src/pages/ChapterMembers.vue
harshtandiya: Question: what if the lead leaves? We don't want a deadlock. 

Also, there's no provision to handle role change. When we create something for role change, we can make this change. 

Won't be doing this in current PR

fossunited/api/chapter.py
RiyaMathew-11: > All shouldn't be a problem, as the chapter's info is public anyway

There are chapter members' info linked to it as well - I think we had the same discussion when we were deciding permissions for Localhosts - and decided to remove `All`?

dashboard/src/components/chapter/AddMember.vue
RiyaMathew-11: > People with anonymous profile still need to be added to the organizer list, right? Should they choose to organize things.

They need to be, by these are private profiles and we've mentioned - profile (in connection to it - data) would be unavailable (that includes name too?) - because when we try to visit a private profile it goes to 404, hence entire data is completely blocked to access as of now.


With the listing - we are exposing it on our side. If the listing isn't there - the private user profiles would share (usernames) it on their own to add them in - which is fine.

Hence I just couldn't relate with this visible listing practice on a dashboard.

dashboard/src/pages/ChapterMembers.vue
RiyaMathew-11: The same case was mentioned in this issue before - #334 4th point

I think we can handle the button visibility in UI now and when #334 is fixed the Role change can be done. Besides the lead won't leave without notice. - as they cannot leave themselves in the system, as we don't have self Leave CTA yet. 

We would have to remove and assign a new lead from the backend until #334 is fixed.

fossunited/chapters/doctype/foss_chapter/foss_chapter.py
RiyaMathew-11: > No, the code is not executed on administrator level on server side. I tried it, it will throw a Permission error on User doctype.

That sort of reverts me back to the first doubt - can ignore_permissions not handle this?

Or is it not at all applied to the Roles workflow ?

Similar seem to be the case here I guess - https://discuss.frappe.io/t/ignore-permissions-to-edit-user-data/85619 but solution isn't very clear.

dashboard/src/components/chapter/AddMember.vue
RiyaMathew-11: We needn't make a change in this PR now too - but keep it as a point of discussion, as I would like to hear more thoughts on this. 

Could go with majority opinion on this also.

fossunited/chapters/doctype/foss_chapter/foss_chapter.py
harshtandiya: Again, as i already told, add_roles and remove_roles method does not work with ignore permissions. This is not a save operation or a create operation.

dashboard/src/pages/ChapterMembers.vue
harshtandiya: Not taking up in this PR. Will make changes with the feature of changing roles

RiyaMathew-11: Approving the PR since the issue already exists on that note.

RiyaMathew-11 (approved):
Approving the PR since it's functionally okay. With doubts existing. 

Resolve conversations and the PR can be merged.

fossunited/chapters/doctype/foss_chapter/foss_chapter.py
RiyaMathew-11: As referred with @maharshivpatel - the approach seems off, tho it seems our only visible option now. 

It would be good to see if there are alter options.

surajshetty3416: Just add a tweaked version of utility function like following so that everyone is aware of what's going on instead of temporarily making a user Admin? 

with this
- You'll get the history of who made the changes if required
- And you don't have to worry about proper session restore

```python
def remove_user_roles(user_name, *roles):
	user = frappe.get_doc("User", user_name)
	existing_roles = {d.role: d for d in user.get("roles")}
	for role in roles:
		if role in existing_roles:
			user.get("roles").remove(existing_roles[role])
	user.save(ignore_permissions=True)
```

harshtandiya: brilliant @surajshetty3416 , this works!

Issue #485, opened 2024-07-20
[FEATURE] New "doctype_ids" module


rahulporuri:
### Feature description

The identifiers of the various doctypes in the codebase are distributed and there is no central place where they are documented. Because of this, we have a case where the doctype id strings are being duplicated in a number of different locations, causing potential issues e.g. trivial spelling mistakes. It will be useful to have a `doctype_ids` (naming things is hard) module where we document this information. The rest of the `fossunited` codebase can then import from this module.

RiyaMathew-11:
I think the issue is same as #446?

rahulporuri:
sigh. yes. they're duplicates. i forgot about the previous issue :/

mngshm:
The older PR was pretty messed up and it became even more complicated because of the merge conflicts. I am reopening a fresh PR today.

rahulporuri:
> The older PR was pretty messed up and it became even more complicated because of the merge conflicts. I am reopening a fresh PR today.

Thanks for the headsup @Mangeshrex .

mngshm:
Format: <constant_variable = <doctype_name> 

#### Chapters Module 

- [x] `CHAPTER` = "FOSS Chapter"
- [x] `CHAPTER_EVENT` = "FOSS Chapter Event"
- [ ] `CHAPTER_EVENT_MEMBER` = "FOSS Chapter Event Member" 
- [ ] `EVENT_PARTICIPANT` = "FOSS Chapter Event Participant" 
- [ ] `CHAPTER_LEAD` = "FOSS Chapter Lead Team Member" 
- [ ] `EVENT_COMMUNITY_PARTNER` = "FOSS Event Community Partner"
- [x] `EVENT_RSVP` = "FOSS Event RSVP" 
- [x] `RSVP_SUBMISSIONS` = "FOSS Event RSVP Submissions"
- [ ] `EVENT_TYPE` = "FOSS Event Type"
- [ ] `TICKET` = "Conference Ticket" 

### FOSS Profiles Module

- [x] `USER_PROFILE` = "FOSS User Profile"
- [ ] `USER_EDUCATION` = "FOSS User Profile Education" 
- [ ] `USER_WORK_EXPERIENCE` = "FOSS User Profile Work Experience" 
- [ ] `USER_SKILL` = "FOSS User Skill" 
- [ ] `USER_SKILL_MULTISELECT` = "FOSS User Skill Multiselect" 
[...]

PR #538, opened 2024-08-09
Replacing doctype names with constants in FOSS United module

mngshm:
- [ ] 🍕Feature
- [x] ⚙️Chore
- [ ] 🐛Bug Fix
- [ ] 🎨Style
- [x] 🧑‍💻Refactor
- [ ] 📕Documentation
- [ ] 🏎️Optimization

## Description
This PR replaces all the doctype names with constants from the FOSS United Module


fossunited/api/dashboard.py
RiyaMathew-11, line 1: What does replace constant in the commit message mean ? Replace which constant ?

fossunited/doctype_ids.py
RiyaMathew-11, line 1: Better commit message maybe ? The changes made is not relevant to the commit message

fossunited/api/dashboard.py
mngshm: `(doctype constant)` is the domain. I am adopting to `<convention> (domain): <message>` style of commits as that makes more sense to me. 

The **commit title** simply means that it will be replacing the "FOSS Event CFP" with the constant (from doctype_ids module) with the defined constant `EVENT_CFP.` 

Please check the issue #485 if you don't have much clarified context about this PR or the whole doctype id to constant thing.

RiyaMathew-11: @Mangeshrex I understood your practice. And also have full context about the PR.

But here I am referring specifcally to this particular commit message: 
[`chore: replace constant and add rest of the constants`](https://github.com/fossunited/fossunited/pull/538/commits/1a788c7e06885df9cb3bd9a517a355c40ac384f6) says just about replace constant and add rest of constants - which is where I was initiallty confused about - which constant are you referring to. It's unclear from the message.

mngshm: Capture-Aug-14-11-04-13

The part of commit message referring to replacing the doctype names with constants. 

Capture-Aug-14-11-05-05

The part of commit message referring to adding the rest of the constants to the files.

fossunited/doctype_ids.py
mngshm: Capture-Aug-14-11-20-49

Just fpushed a commit to amend the previous commit message. I hope this conversation can be resolved now.

fossunited/fossunited/forms.py
```diff
@@ -2,7 +2,7 @@
 
 import frappe
 
-from fossunited.doctype_ids import USER_PROFILE
+from fossunited.doctype_ids import CUSTOM_ANSWERS, USER_PROFILE
```
rahulporuri, line 5: :red_circle: why we immediately need to fix ruff 

```suggestion
from fossunited.doctype_ids import CUSTOM_ANSWER, USER_PROFILE
```

RiyaMathew-11 (approved):

fossunited/doctype_ids.py
```diff
@@ -12,3 +12,23 @@
...
+EVENT_FIELD = "FOSS Event Field"
+EVENT_SCHEDULE = "FOSS Event Schedule"
+EVENT_SPONSOR = "FOSS Event Sponsor"
+GLOBAL_CFP_REVIEWER = "FOSS Global CFP Review Member"
+GLOBAL_REVIEW_SETTINGS = "FOSS Global CFP Review Settings"
+PROJECT_GRANT = "FOSS Project Grant"
+TEAM = "FOSS United Team"
+FOSSUNITED_SETTINGS = "FOSSU Settings"
+INDUSTRY_PARTNERS = "Industry Partners"
+ORGANIZATION = "Organization"
```
rahulporuri: it doesn't look like they are all being used but that's okay :shrug:

rahulporuri (approved):
LGTM

fossunited/fossunited/forms.py
mngshm: could you please check the conversation on mattermost regarding ruff? I assume you missed it.

fossunited/doctype_ids.py
mngshm: Yes, they are not used. But I added them in case if we have to use these constants in the near future.

RiyaMathew-11: `TEAM`, `INDUSTRY_PARTNERS` are used in web templates - I assumed we are not replacing constants in web templates in this PR.

`FOSSU_SETTINGS` is actually used in the signup workflow - which is a bit outside the scope of fossunited app in general. It's also referred [here](https://github.com/fossunited/fossunited/blob/959f3860440aae6a9fb4262e274268bd9c7c96a2/fossunited/fossunited/utils.py#L164).
Screenshot 2024-08-14 at 12 21 30 PM


`ORGANIZATION` is something that I think is not in use now -I couldn't find any relevant ones on a first search - we could check again and see.

mngshm: We are only replacing in the py. please correct me if it is getting used anywhere. Locally, I had tried finding their use in any of the py files multiple times through `grep`, `xargs` and also at the end with vscode.

RiyaMathew-11: Yep - web templates are html pages, hence not relevant to this PR, if we are not focusing on files other than `.py`

However, FOSSU Settings seem in `utils.py`, which is already changed too.

mngshm:
image

Gave up.  While amending, I force pushed to wrong repo, checked the head on the wrong branch and pushed the head to this branch. Which made it merge upstream, and unnecessary changes came in (for ex: deletion of files, and HACKATHON Constant changes which are already merged. 

I'll redo this PR asap and get it merged. Apologies.

PR #544, opened 2024-08-12
fix(cfp): user nonetype error in cfp pages

harshtandiya:
- [x] 🐛Bug Fix

## Description

- Reviews made through the dashboard were not capturing the 'email' field for FOSS Event CFP Review child doc. This field was used to render the picture of the Reviewer. With a blank email, the `User Nonetype` error was being thrown.
- Fixed it so emails are autofetched from the User Profile link field in CFP Review child doc.
- Will run a script on console via desk so that all other reviews are fixed.


mngshm (approved):
The PR Looks fine, Approving it.

rahulporuri:
It would have been good to add a regression test in this PR

PR #562, opened 2024-08-19
fix(chapter): fix chapter members comparison bug

### LINKED ISSUE #546: [BUG] Every chapter member loses role on member removal
### Describe the bug

Every member of the chapter seems to be losing the role of `Chapter Team Member` even if a single member is removed

harshtandiya:
Fix a bug in the `FOSS Chapter` doctype where the comparison of chapter members was not done correctly. This bug caused issues when checking if a member was already a team member or not; resulting in role removal of all the members of chapter. Now, the comparison is done using the `chapter_member` attribute of each member (which maps to profile id), ensuring accurate results.

- [x] 🐛Bug Fix

## Related Issues & Docs
closes #546

### DIFF
--- a/fossunited/chapters/doctype/foss_chapter/foss_chapter.py
+++ b/fossunited/chapters/doctype/foss_chapter/foss_chapter.py
@@ -25,7 +25,7 @@ class FOSSChapter(WebsiteGenerator):
         chapter_lead: DF.Link | None
         chapter_members: DF.Table[FOSSChapterLeadTeamMember]
         chapter_name: DF.Data
-        chapter_type: DF.Literal["City Community", "FOSS Club", "Conference"]  # type: ignore
+        chapter_type: DF.Literal["City Community", "FOSS Club", "Conference"]
         city: DF.Link | None
         country: DF.Link | None
         email: DF.Data
@@ -44,7 +44,7 @@ class FOSSChapter(WebsiteGenerator):
 
     # end: auto-generated types
     def before_insert(self):
-        self.set_member_roles()
+        self.handle_member_addition()
 
     def validate(self):
         self.make_city_name_upper()
--- a/fossunited/chapters/doctype/foss_chapter/foss_chapter.py
+++ b/fossunited/chapters/doctype/foss_chapter/foss_chapter.py
@@ -25,7 +25,7 @@ class FOSSChapter(WebsiteGenerator):
         chapter_lead: DF.Link | None
         chapter_members: DF.Table[FOSSChapterLeadTeamMember]
         chapter_name: DF.Data
-        chapter_type: DF.Literal["City Community", "FOSS Club", "Conference"]
+        chapter_type: DF.Literal["City Community", "FOSS Club", "Conference"]  # type: ignore
         city: DF.Link | None
         country: DF.Link | None
         email: DF.Data
@@ -79,7 +79,7 @@ class FOSSChapter(WebsiteGenerator):
         if not prev_doc:
             return
         for member in prev_doc.chapter_members:
-            if member not in self.chapter_members:
+            if member.chapter_member not in [d.chapter_member for d in self.chapter_members]:
                 if self.member_of_other_chapter(member):
                     continue
                 user = frappe.db.get_value(USER_PROFILE, member.chapter_member, "user")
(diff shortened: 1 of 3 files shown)


fossunited/chapters/doctype/foss_chapter/foss_chapter.py
```diff
@@ -25,7 +25,7 @@ class FOSSChapter(WebsiteGenerator):
         chapter_lead: DF.Link | None
         chapter_members: DF.Table[FOSSChapterLeadTeamMember]
         chapter_name: DF.Data
-        chapter_type: DF.Literal["City Community", "FOSS Club", "Conference"]
+        chapter_type: DF.Literal["City Community", "FOSS Club", "Conference"]  # type: ignore
```
rahulporuri: it's not obvious why this type hint is being ignored.

rahulporuri (dismissed):
@harshtandiya please add a regression unit test in this PR. It looks like we don't have any blocker PRs at the moment so please let's take the time to fix this problem right once and for all.

fossunited/chapters/doctype/foss_chapter/foss_chapter.py
harshtandiya: oof, this was not supposed to be pushed. was experimenting

RiyaMathew-11 (commented):
Functionally right. PR can be approved once after the unit-test addiion

mngshm (commented):
Looks legitimate, will wait for the tests to be added.

RiyaMathew-11 (commented):
@harshtandiya CI tests are failing on member assignment and de-assignment tests. Please check on that.

Let's also use this chance to add Faker to [pyproject.toml](https://github.com/fossunited/fossunited/blob/develop/pyproject.toml) as discussed before.

fossunited/chapters/doctype/foss_chapter/test_foss_chapter.py
```diff
@@ -1,8 +1,95 @@
...
+            }
+        )
+        chapter.insert()
+
+        if not frappe.db.exists("Role", "Chapter Team Member"):
+            frappe.get_doc({"doctype": "Role", "role_name": "Chapter Team Member"}).insert(ignore_permissions=True)
+        if not frappe.db.exists("Role", "Chapter Lead"):
+            frappe.get_doc({"doctype": "Role", "role_name": "Chapter Lead"}).insert(ignore_permissions=True)
+
+        lead_profile = frappe.get_doc("FOSS User Profile", {"user": "test@example.com"})
```
RiyaMathew-11, line 37: This test@example.com is assumed to always exist right ?

fossunited/chapters/doctype/foss_chapter/test_foss_chapter.py
```diff
@@ -1,8 +1,95 @@
...
+        frappe.delete_doc("FOSS Chapter", self.chapter.name)
+
+    def test_role_assignment_on_create(self):
+        # Given a chapter
+        chapter = self.chapter
+
+        # When the chapter was created, a lead was assigned to it.
+        # Then the lead should have the role of 'Chapter Lead' and 'Chapter Team Member'
+        user = frappe.db.get_value("FOSS User Profile", chapter.chapter_members[0].chapter_member, "user")
+        has_role = frappe.db.exists("Has Role", {"role": "Chapter Team Member", "parent": user})
```
RiyaMathew-11, line 53: I am a bit confused about 'Has Role' mention in frappe.db.exists() call. Haven't seen the usage before. Please explain this ?

fossunited/chapters/doctype/foss_chapter/test_foss_chapter.py
RiyaMathew-11: When I ran the tests - it failed for me first, as the users didnt exist. It might be good if we create dummy users here too? For the tests and delete them in tearDown ?

RiyaMathew-11 (commented):
Made a couple of comments @harshtandiya - Please resolve them and we are good to go.

fossunited/chapters/doctype/foss_chapter/test_foss_chapter.py
harshtandiya: Test users are created by frappe if tests are enabled on your bench. They are always created, unless you have explicitly deleted them

fossunited/chapters/doctype/foss_chapter/test_foss_chapter.py
harshtandiya: Has Role is a child table in Frappe which houses different roles which are linked to a parent of `User`. 

If you have a `Chapter Team Member` role. then `Has Role` will have an entry with `parent` as your user.

RiyaMathew-11 (approved):

PR #571, opened 2024-08-21
chore(event): better data display of speakers in event page

harshtandiya:
- [x] ⚙️Chore

## Description

- Display Name from submission doctype rather than the profile.
- Remove `a` tag and hover effects
- Show designation and company instead of talk title.

image

### DIFF
--- a/fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
+++ b/fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
@@ -215,17 +215,17 @@ class FOSSChapterEvent(WebsiteGenerator):
                 "event": self.name,
                 "status": "Approved",
             },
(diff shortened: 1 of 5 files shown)


RiyaMathew-11 (approved):

RiyaMathew-11:
A few thoughts - in case folks would want to know more about the speaker in a way - we could add some reference link of the speaker in future. 

- FOSS Profile if exists
- Social media links if given

rahulporuri:
> FOSS Profile if exists

^ . The social media links can be accessed from here so I wouldn't add anything else.

RiyaMathew-11:
> > FOSS Profile if exists
> 
> ^ . The social media links can be accessed from here so I wouldn't add anything else.

Edge case for those speakers who don't have a profile. Like the invited proposals we did for IndiaFOSS

Issue #577, opened 2024-08-28
give icon/svg attributions in codebase


harshtandiya:
### Issue description

The codebase uses a lot icons, majorly from [Tabler](https://github.com/tabler/tabler?tab=MIT-1-ov-file). Along with a few other illustrations.

It would be a good practice to give attribution to the SVGs, Icons and Illustrations used in the code.

Additional: 
- we need to source the illustration attributions from Jeswin, so that we can attribute them correctly

Tangential:
- We can also think about having tabler icons as a dependency, just because we use it so much. [See here](https://tabler.io/docs/icons/vue)

harshtandiya:
done for the most part

Issue #224, opened 2024-04-02
[FEATURE] Calendar Integrations


ansharora28:
### Feature description

Adding a .ics or gcal/ical integration to events and talk schedules

### Use case

 For the talks schedule, adding a .ics integration will help people directly add the talks to their calendar, view it offline, get reminders (maybe volunteers can also move these talks around depending on the schedule,delays etc.)

abhayguptas:
Hi, So the end goal for this is generating a .ics file when someone RSVP for any event and giving the user a button to download that file, right?

harshtandiya:
The goal is to generate and send the ics file in the notification email

mngshm:
Closing this issue as #302 is a similar issue.

PR #584, opened 2024-08-29
feat: `Add to Calendar` button for schedule sessions

### LINKED ISSUE #302: [FEATURE] Calendar integration for schedule page
### Feature description

Create an `Add to Calendar` integration for conference sessions in schedule. 

This is a special request that came from the community in connection with major conferences like IndiaFOSS: 

- An Add to Calendar feature would enable attendees to add those particular talks or sessions to their calendar. 
- Enables reminder for attendance.

~In addition, a calendar integration for events is also needed.~

harshtandiya:
- [x] 🍕Feature

## Description

- Add toggle view for schedules. Now the user has a choice to view the schedule in either `Horizontal` or `Vertical` view.
- Adds functionality to download the .ics file per session of the schedule.

## Related Issues & Docs

closes #302  

[Screencast from 2024-08-29 17-29-31.webm](https://github.com/user-attachments/assets/24254c08-3e26-4e9c-84e2-57624d35aa81)


dashboard/src/components/icons/CalenderAddIcon.vue
```diff
@@ -0,0 +1,24 @@
+<template>
+  <svg
```
rahulporuri, line 2: Just to confirm, the SVG icons also came from Jeswin and will be added to #577 ?

pyproject.toml
```diff
@@ -10,6 +10,7 @@ dynamic = ["version"]
 dependencies = [
     "razorpay~=1.4.2",
     "PyGithub~=2.3.0",
+    "ics~=0.7.2",
```
rahulporuri: tangential comment - the last release was 2+ years ago but I don't know if there is a more up-to-date alternative.

fossunited/api/schedule.py
```diff
@@ -94,3 +96,30 @@ def get_schedule_by_hall(schedule: list) -> dict:
         _schedule[hall].append(session)
 
     return _schedule
+
+
+@frappe.whitelist(allow_guest=True)
+def get_schedule_session_ics(schedule: dict):
```
rahulporuri:
```suggestion
def get_schedule_session_ics(schedule: dict) -> str:
```

fossunited/api/schedule.py
```diff
@@ -94,3 +96,30 @@ def get_schedule_by_hall(schedule: list) -> dict:
...
+    """
+    Download the ics file for the schedule session
+
+    Args:
+        schedule (dict): Schedule session data
```
rahulporuri: add a return too?

fossunited/api/schedule.py
```diff
@@ -94,3 +96,30 @@ def get_schedule_by_hall(schedule: list) -> dict:
...
+    event.name = schedule.get("title")
+    event.begin = timezone.localize(datetime.strptime(f'{schedule.get("scheduled_date")} {schedule.get("start_time")}', "%Y-%m-%d %H:%M:%S"))
+    event.end = timezone.localize(datetime.strptime(f'{schedule.get("scheduled_date")} {schedule.get("end_time")}', "%Y-%m-%d %H:%M:%S"))
+    event.location = f'{schedule.get("hall")} - {event_location}'
+
+    cal.events.add(event)
+
+    ics_content = cal.serialize()
+
+    return ics_content
```
rahulporuri: :sweat_smile: did you test this locally to confirm that it creates a valid ICS file? It should be pretty easy and quick to create a unittest for this function so consider adding one in the PR itself

rahulporuri (approved):
Overall LGTM but like usual, I'm useful when it comes to HTML/CSS/JS/Vue stuff so @RiyaMathew-11 please take a look at the code if you have the time.

I've added a few comments @harshtandiya and it looks like we have time before the schedule is ready so please consider adding a unittest to verify that the ICS files work as expected.

RiyaMathew-11:
@harshtandiya #302 is a valid issue for this PR, needn't have created a duplicate issue.

Edit: Updated PR description with #302 and closed the duplicate one.

fossunited/api/schedule.py
```diff
@@ -94,3 +96,30 @@ def get_schedule_by_hall(schedule: list) -> dict:
...
+        schedule (dict): Schedule session data
+    """
+
+    cal = Calendar()
+
+    timezone = pytz.timezone("Asia/Kolkata")
+
+    event_location = frappe.db.get_value("FOSS Chapter Event", schedule.get("parent"), "event_location")
+    event = Event()
+    event.name = schedule.get("title")
```
RiyaMathew-11: I would suggest to add the event name in the Title too - it currently says about just session title, which can be confusing at some points too. 
Screenshot 2024-08-29 at 9 37 23 PM

fossunited/api/schedule.py
RiyaMathew-11: Tangential - but this method could prove to be generic to add an Add to Calendar Integration for an event as a whole. 

Only if it's decided to make it as an integration in the event page, as referred in #585.

pyproject.toml
RiyaMathew-11: [icalendar](https://pypi.org/project/icalendar/) seems a good alternative ? Extensive and seems well maintained.

fossunited/api/schedule.py
```diff
@@ -94,3 +96,30 @@ def get_schedule_by_hall(schedule: list) -> dict:
...
+    event.begin = timezone.localize(datetime.strptime(f'{schedule.get("scheduled_date")} {schedule.get("start_time")}', "%Y-%m-%d %H:%M:%S"))
+    event.end = timezone.localize(datetime.strptime(f'{schedule.get("scheduled_date")} {schedule.get("end_time")}', "%Y-%m-%d %H:%M:%S"))
+    event.location = f'{schedule.get("hall")} - {event_location}'
+
+    cal.events.add(event)
```
RiyaMathew-11: Could also consider for a reminder option here preferably - preferably default 10 min before, as the event itself could be very extensive. 

Could be done with `event.alarms` I think that comes with the `Event` class.

Getting reminders was also mentioned in the OG issue - #224

dashboard/src/components/icons/CalenderAddIcon.vue
harshtandiya: Missed this one, this is also a Tabler Icon. I will go ahead and add

fossunited/api/schedule.py
harshtandiya: I will create a generic one when I am working on it. Don't see the need to overcomplicate things at this point in time tbh

fossunited/api/schedule.py
harshtandiya: I did test it locally, and it did create valid ics files(shown in video).

Will try adding a unittest, that could be helpful

pyproject.toml
harshtandiya: I can try replacing this with icalender if that's what we want?

RiyaMathew-11: I would suggest a better alternative would be to not take it to the backend at all. 

There is a time-factor involved. When I click the button, there is atleast a 2-3 s delay without any loading state in the UI - a bit confusing at first. This covers the entire cycle of sending request -> processing -> fetching response.

We could potentially handle this on the client side - there should be JS libraries that support calendaring. The entire data needed is there in the frontend. Handling the generation of an ics on the client side would make it faster too.

fossunited/api/schedule.py
RiyaMathew-11: Refer this too - https://github.com/fossunited/fossunited/pull/584#discussion_r1736895097

fossunited/api/schedule.py
RiyaMathew-11: Refer this - https://github.com/fossunited/fossunited/pull/584#discussion_r1736895097

pyproject.toml
RiyaMathew-11: This might be a helpful JS library here to use - [ics](https://www.npmjs.com/package/ics).

dashboard/src/components/schedule/ScheduleView.vue
```diff
@@ -0,0 +1,61 @@
...
+    }"
+  >
+    <div
+      v-for="(sessions, hall) in schedule"
+      :key="hall"
+      :class="{
+        'min-w-[720px] flex-shrink-0': view === 'horizontal',
+      }"
+    >
+      <SessionListHeader
```
RiyaMathew-11, line 16: There are built-in HTML tags that handle this particular use case of collapsible components: [`details`, `summary`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details). 

Styling can be done as per how we want. The main advantages are: 

1. We needn't monitor the state of the Vue components, as it's builtin with the tag. 
2. It enhances accessibility - as it's detected well by screen-readers etc.

Which makes it relevant as we are trying to make IndiaFOSS accessible as mush as possible. 

This could also be taken as a potential refactor in future.

RiyaMathew-11: Tried experimenting with it as whole with 0 style add-ons: 

```
<details>
        <summary>
          {{ hall }}
          <!-- <SessionListHeader
          :title="hall"
          :collapsible="isCollapsible"
          @collapse-hall="toggleCollapse(hall)"
          /> -->
        </summary>

        <SessionList
          v-if="!isCollapsed(hall)"
          :sessions="sessions"
          :view="view"
        />
</details>
```

Result: 
Screenshot 2024-08-29 at 11 45 49 PM

RiyaMathew-11 (commented):
LGTM functionally - added a bunch of comments in between some suggesting potential add-ons and refactors. Do check them out @harshtandiya. 

@rahulporuri would be nice to have your thoughts on some relevant ones too. - https://github.com/fossunited/fossunited/pull/584#discussion_r1736895097, https://github.com/fossunited/fossunited/pull/584#discussion_r1736918729

RiyaMathew-11 (approved):

pyproject.toml
harshtandiya: I genuinely don't feel like this should be something we should throw away at client devices to handle 😅 

Moreover, I feel data can be fetched more easily (event title, event date, event location) in the server side using Frappe apis, and easily added to ics files

RiyaMathew-11: > I genuinely don't feel like this should be something we should throw away at client devices to handle 😅

Can you please explain why? 

IMO the data is literally there in the frontend - why add an additional server hop to this ? Event date, Event title are also there - fetched for the header ? Event location can also be something that's fetched alongside these event details. That's a single time event.

While in the server-side - you are making that step redundant too, fetching the event details each time the API is called.

Btw, do feel free to ignore if not relevant.

harshtandiya: > Can you please explain why?

Not really, I just find that method to be something which i find to be quicker and easier to implement. I could be wrong too. Please start a new issue/discussion to discuss what method/technologies would be better to use in this regard. Not addressing this in this PR

dashboard/src/components/schedule/ScheduleView.vue
harshtandiya: The Schedule List header is collapsible sometimes, and other times its not. Depends on the view. Can try experimenting this in future

RiyaMathew-11: > header is collapsible sometimes, and other times its not

This can be handled via tag attributes - check documentation for more details. This needn't be implemented now also, as I've mentioned as: 

> This could also be taken as a potential refactor in future.

pyproject.toml
rahulporuri: @harshtandiya, I do agree with @RiyaMathew-11  here about doing this on the client side instead of the server side.

The only reason why we're being forced to make an API call is because `FOSSEventSchedule` does not contain the `event_location` and we should just update the doctype to include the `event_location` attribute, set to the event location from the `FOSSChapterEvent` doctype

Once we do that, the client has all of the information in the session object and it doesn't need to make an API call at all, at which point we can just use the `ics` npm package to create the ICS file.

Along with Riyas' argument about better performance because of one less API call, I advocate for doing this on the client side because it reduces our API surface area. It's a low risk API but it's still one more API endpoint and anytime we can get away with adding less code, we should IMHO.

@harshtandiya can we please make this change in this PR? On the bright side, you can ignore my comment about the unittest because the endpoint itself won't exist :D

pyproject.toml
```diff
@@ -10,8 +10,6 @@ dynamic = ["version"]
 dependencies = [
     "razorpay~=1.4.2",
     "PyGithub~=2.3.0",
-    "ics~=0.7.2",
-    "Faker~=27.0.0",
```
rahulporuri: it looks like we accidentally removed `Faker`?

rahulporuri:
Still LGTM, except for a minor issue with a dependency that might have been accidentally removed. Thank you very much for making the changes @harshtandiya !

pyproject.toml
harshtandiya: Fixed

Issue #582, opened 2024-08-28
Make reusable functions as general helper functions


RiyaMathew-11:
### Issue description

Recently in one of the PRs - #572, came across a function, written in the narrow scope of a page, that could potentially be reusable. It was considered to [make as a helper function in a separate PR](https://github.com/fossunited/fossunited/pull/572#discussion_r1734350607). 

There could also be other functions / instances of the same note, in the codebase in general. Generally we should think of adopting the practice of making these utility functions as helpers.

rahulporuri:
closing as not planned - we will refactor as and when the code is modified

PR #572, opened 2024-08-21
feat: event schedule pages

harshtandiya:
## What type of PR is this? (check all applicable)
- [x] 🍕Feature

## Description
This PR deals with adding web page for schedules, which will be a standard for events hosted on platform.

- Refactor `title` field, and adds fields such as `category`, `other_category`, `hall` and `speakers` for the schedule child doctype. 

The category field will be displayed on the session cards.   

The `hall` field will have the hall name where the talk takes place.

The `speakers` field is a `JSON` type field in which multiple speakers can be stored in array type item with key "speakers".
```
{
    "speakers": [],
}
```

- add `hall_options` field in event doctype. This will have list of halls in the venue.
- `event_logo` field added in the Event doctype
- Add `schedule_page_description` field in Event doctype. This enables chapter members to have a custom description in the schedule page if required. If left empty, a standard generic description is shown in its place.
- Breadcrumb components updated to support external urls. This can help us redirect directly to the web page of any external event directly.
- FossUnitedLogo and FossClubLogo are added as vue components. Converting SVGs to vue components help to add additional DOM properties/event handlers. Also, styling with CSS is easier.
-  Add Schedule Header component, which consists of the schedule image, title and description.
- Add ScheduleDateToggle component, which has the different dates of schedules as toggle buttons. This helps in toggling between the schedules of different dates of a multi-day event.
- SessionListHeader component acts as an accordion for SessionList component. `title` prop is passed to it, which is usually the name of the 'hall'.
- Session List component is the parent component which renders the schedule blocks for a list of sessions.
- Schedule.vue page added which houses all the above components to provide with the schedule page.

- `get_event_schedule` api returns the schedule of an event arranged in a dict with the keys being dates of session, and sessions list as values. These session lists are further arranged in a dict with keys being the 'halls' and values being sessions. 

Ex:
```
Returns:

{
    "07/09/2024": {
        "Audi 1" : [{session1}, {session2}],
        "Audi 2": [{session1}]
        }
    },
    "08/09/2024":{
        "Open Space": [{session1}]
    }
}

```

## Related Issues & Docs

closes #540 

### Desktop
[Screencast from 2024-08-27 18-58-24.webm](https://github.com/user-attachments/assets/1e4079bb-e4ac-4909-8394-af8bb22d41d3)

### Mobile
[Screencast from 2024-08-27 18-59-42.webm](https://github.com/user-attachments/assets/e4682f5f-f1bc-4374-9d1b-2d12b565d562)

Issue #529, opened 2024-08-07
[QUESTION] Should User Profiles duplicate functionality provided by existing social media accounts


rahulporuri:
### Question or issue

At the moment, a FOSS United User Profile has the following

- About
- Work
- Resume (Experience and Education)
- Projects

The User Profile also allows the user to provide links to Social Media accounts e.g. X/Twitter, LinkedIn, YouTube, GitHub, GitLab, Mastadon, etc

- LinkedIn has become the defacto place for people to share their "Resume" i.e. work experience and educational qualifications. 
- GitHub, GitLab, and other Git source control websites are the defacto places to list "Projects"

What value do the "Resume" and "Projects" tabs provide that the existing social media accounts don't. In my personal and humble opinion, Users are less likely to choose to duplicate information that they have shared on other Social Media accounts. GitHub also enables users to "Pin" and highlight specific repositories so what additional value does the "Projects" tab provide? FOSS United-specific technical projects/contributions?

Are we supporting the small percentage of Users that don't use LinkedIn via this functionality?

### Additional context

Personal opinion - the FOSS United User Profiles should enable FOSS United-specific information sharing. The users can provide their FOSS United community and technical contributions on other social media platforms but the root/source information remains with us - so that is what we should focus on, and not on duplicating existing social media accounts. The long-term cost of maintaining this additional functionality is non-zero and we should seriously consider whether we want to make any changes to this code and/or keep this code around.

PR #589, opened 2024-09-01
refactor: redesign talk proposal page

### LINKED ISSUE #512: Redesign Talk Proposal Blocks
### Issue description

The talk proposal blocks are not at all innovative when it comes to design right now. More than anything, they seem to confuse people.

Few suggestions to make them better:
- Redesign them as per this suggested design by Jeswin - https://www.figma.com/design/Wo7sfsCLpyHgU0iIcrFeiR/IF24-webUI?node-id=0-1&t=z0hil6pvRCS3ozu4-1
- Add a search and filter function in the /all page for talk proposals

RiyaMathew-11:
- [x] 🎨Style
- [x] 🧑‍💻Refactor
- [x] 🏎️Optimization

## Description

This PR contains the redesigned `All proposals` page as given in this [UI design](https://www.figma.com/proto/Wo7sfsCLpyHgU0iIcrFeiR/IF24-webUI?node-id=228-514&node-type=FRAME&m=dev&scaling=contain&content-scaling=fixed&page-id=0%3A1&t=1rlmzhdlh9teqCK6-1).

### Features included: 

  - Search by proposal name
  - Filter by `Session Type`, `Status`

---

- Two new UI components have been added - `AllProposalsBlock` and `AllProposalsBanner`
  - The `EventHeader` component from schedules have been reused.
  - Used styled variant for search and wrote a component `ThemedSelectBlack` for filter buttons

- An additional text field - Proposal page description is added to `FOSS Chapter Event` doctype. 
 
- The workflow makes use of 2 whitelisted APIs - as default methods like `get_list` and `get_value` aren't whitelisted for guest users and this page is a public page.  

  - `get_event_from_permalink() is reused, while `get_event_proposals` is added to fetch proposal lists based on conditions applied like anonymise CFPs etc.,

- Removed `Recent Submissions` section from Talk Proposal Section of Event Page, and added a button component to redirect to proposal page.

## Related Issues & Docs

- Closes #512 

## Checklist
- [x] ~The code follows the project's coding standards~.

Screenshot 2024-09-05 at 10 17 40 PM

all-proposals

Screenshot 2024-09-02 at 8 58 36 PM

Screenshot 2024-09-05 at 10 17 56 PM

## Additional context

- The session types and status right now aren't fetched from backend, could do that separate so that whenever more session-types are updated in doctype, they could be dynamically fetched

- The 'like feature' for proposals isn't implement upfront in the list of proposals - it just shows the count of likes now. This could also be implemented in future. 
  - Sort by `number of likes` can be implement after this too.

- Added a helper function `createAbsoluteUrlFromRoute ` that can be re-used in breadcrumb links etc., - for a link to be opened in the same tab. As referenced in #582


harshtandiya:
please try to extract code into components

RiyaMathew-11:
> please try to extract code into components

There are further design changes @harshtandiya that Jeswin made now to make generic components more uniform - will be doing that as components. Please check the design file for changes.

This PR isn't fully ready yet. 

Edit: Have extracted components for re-usable cases.

dashboard/src/components/Breadcrumb.vue
```diff
@@ -8,7 +8,7 @@
       <router-link v-if="item.route" class="hover:text-gray-700 transition-colors" :to="item.route">
         {{ item.label }}
       </router-link>
-      <a v-else-if="item.link" :href="item.link" target="_blank" class="hover:text-gray-700 transition-colors">
```
harshtandiya, line 11: Why

dashboard/src/pages/events/AllProposals.vue
```diff
@@ -0,0 +1,252 @@
...
+import { computed, ref } from 'vue'
+import { useRoute } from 'vue-router'
+import { createResource, usePageMeta, LoadingText } from 'frappe-ui'
+import { redirectRouteToSameWindow } from '@/helpers/utils'
+import Header from '@/components/Header.vue'
+import Breadcrumb from '@/components/Breadcrumb.vue'
+import AllProposalsBanner from '../../components/proposals/AllProposalsBanner.vue'
+import ProposalBlock from '../../components/proposals/ProposalBlock.vue'
+import EventHeader from '../../components/schedule/EventHeader.vue'
+import ChevronDown from '../../components/icons/ChevronDown.vue'
```
harshtandiya: nitpicky: would be nice to use `@/` here. You dont have to strictly address this

dashboard/src/pages/events/AllProposals.vue
```diff
@@ -0,0 +1,252 @@
...
+
+const cfp = createResource({
+  url: 'fossunited.api.dashboard.get_cfp_details_from_eventname',
+  makeParams() {
+    return {
+      eventname: event.data.event_name,
+      fields: ['route', 'proposal_page_description'],
+    }
+  },
+})
```
harshtandiya: I think this can be replaced with the standard `frappe.client.get` or `frappe.client.get_value` api method, as the FOSS Event CFP has permissions of `All` set to read

dashboard/src/pages/events/AllProposals.vue
```diff
@@ -0,0 +1,252 @@
...
+        'status',
+      ],
+    }
+  },
+  auto: false,
+  onSuccess(data) {
+    console.log(data)
+    proposalLikes.fetch()
+  },
+})
```
harshtandiya: This can just be a single method where you take the Proposal data + likes data in them. 

The function can return something like

```
[
{ proposal 1 details, name, title, likes count},
{ proposal 2 details, name, title, likes count},
{ proposal 3 details, name, title, likes count}
]
```

dashboard/src/pages/events/AllProposals.vue
```diff
@@ -0,0 +1,252 @@
...
+
+const proposalLikes = createResource({
+  url: 'fossunited.api.proposal.get_likes',
+  makeParams() {
+    const proposalIDs = proposals.data?.map((proposal) => proposal.name) || []
+    return {
+      proposals: proposalIDs,
+    }
+  },
+})
```
harshtandiya: Again, this seems unnecessary, all of this data should come with the proposal data in a single api endpoint.

Making an api with that in mind can be helpful

fossunited/fossunited/doctype/foss_event_cfp_submission/templates/foss_event_cfp_submission.html
```diff
@@ -9,7 +9,7 @@
             <span>{{ event.event_name }}</span>
         </a>
         <i class="ti ti-chevron-right"></i>
-        <a href="/{{event.route}}/cfp/all">
```
harshtandiya, line 12: We should add a redirect to /all page, so that it redirects to the dashboard page. 

See for example: schedule.py in www folder

fossunited/chapters/doctype/foss_chapter_event/templates/foss_chapter_event.html
harshtandiya, line 1: A lot of unnecessary code formatting. I cannot find what was changed

fossunited/api/proposal.py
```diff
@@ -0,0 +1,22 @@
...
+    Returns:
+        dict: with proposal id as key and count of likes as values
+    """
+
+    proposalLikes = dict()
+
+    for proposal in proposals:
+        proposalLikes[proposal] = len(get_doc_likes("FOSS Event CFP Submission", proposal))
+
+    return proposalLikes
```
harshtandiya: Again, i find this to be unneccesary

dashboard/src/pages/events/AllProposals.vue
```diff
@@ -0,0 +1,252 @@
...
+              class="absolute z-10 mt-1 w-full bg-white border rounded-sm shadow-lg"
+            >
+              <div
+                v-for="type in sessionTypes"
+                :key="type"
+                @click="selectSessionType(type)"
+                class="px-4 py-2 hover:bg-gray-100 cursor-pointer"
+              >
+                {{ type }}
+              </div>
```
harshtandiya: Why don't you just use a Select with v-model? this seems like a whole lot of code that can be much simpler.

Select from frappeui would work wonderfully here afaik

For implementation you can look in following files: ProposalList.vue, TicketList.vue

dashboard/src/pages/events/AllProposals.vue
```diff
@@ -0,0 +1,252 @@
...
+              class="absolute z-10 mt-1 w-full bg-white border rounded-sm shadow-lg"
+            >
+              <div
+                v-for="status in statuses"
+                :key="status"
+                @click="selectStatus(status)"
+                class="px-4 py-2 hover:bg-gray-100 cursor-pointer"
+              >
+                {{ status }}
+              </div>
```
harshtandiya: same as above

dashboard/src/helpers/utils.js
```diff
@@ -5,3 +5,7 @@ export const truncateStr = (title, len) => {
 export const redirectRoute = (route) => {
   window.open(document.location.origin + '/' + route, '_blank')
 }
+
+export const redirectRouteToSameWindow = (route) => {
+  return window.location.origin + '/' + route
+}
```
harshtandiya: this function can be named better. The name misleads with what it does. 

`getRedirectLink` or something would be better

Highly recommended to read: https://gist.github.com/18alantom/d9f0565c0f42d6a71311d4a3093a1331#file-code_readability-md

dashboard/src/components/proposals/ProposalBlock.vue
```diff
@@ -0,0 +1,80 @@
+<template>
+  <div v-if="proposal && proposalLikes != null" class="flex justify-between">
+    <div class="flex flex-col gap-3 justify-between">
+      <h3 class="text-xl font-semibold">
+        <a
+          :href="redirectRouteToSameWindow(proposal.route)"
```
harshtandiya: Again, function name is misleading

dashboard/src/components/proposals/ProposalBlock.vue
```diff
@@ -0,0 +1,80 @@
...
+          class="hover:text-green-500 transition-colors duration-300 cursor-pointer"
+        >
+          {{ proposal.talk_title }}
+        </a>
```
harshtandiya: Also, i dislike the green color on hover 😅 This is nitpicky, but it seems so outdated. We can have an underline on hover maybe

dashboard/src/components/proposals/ProposalBlock.vue
```diff
@@ -0,0 +1,80 @@
...
+          }}</span>
+          <div>
+            <span
+              class="px-2 py-1 rounded-sm md:hidden"
+              :class="getStatusClass(proposal.status)"
+              >{{ proposal.status.toUpperCase() }}</span
+            >
+          </div>
+        </div>
+        <div v-if="proposal.status == `Approved`" class="flex text-sm">
```
harshtandiya: The organizers can keep the CFPs non-anonymous as well. In that case we should be showing the details of the speaker, irrespective of whether it is Approved or Rejected

dashboard/src/components/proposals/ProposalBlock.vue
```diff
@@ -0,0 +1,80 @@
...
+      <div class="flex items-center gap-1 bg-gray-200 px-1.5 py-1 rounded-sm">
+        <LikesIcon />
+        <span class="text-xs font-semibold">{{ proposalLikes }}</span>
+      </div>
+      <div class="mt-2">
+        <span
+          class="px-2 py-1 rounded-sm text-xs font-semibold hidden md:block"
+          :class="getStatusClass(proposal.status)"
+          >{{ proposal.status.toUpperCase() }}</span
+        >
```
harshtandiya, line 45: Am i seeing this twice? please confirm this was not a mistake

dashboard/src/components/proposals/ProposalBlock.vue
```diff
@@ -0,0 +1,80 @@
...
+        <div class="flex items-center text-xs font-semibold">
+          <span class="bg-gray-200 px-2 py-1 text-gray-600 rounded-sm mr-2">{{
+            proposal.session_type.toUpperCase()
+          }}</span>
+          <div>
+            <span
+              class="px-2 py-1 rounded-sm md:hidden"
+              :class="getStatusClass(proposal.status)"
+              >{{ proposal.status.toUpperCase() }}</span
+            >
```
harshtandiya, line 24: Twice

dashboard/src/pages/events/AllProposals.vue
```diff
@@ -0,0 +1,252 @@
...
+        'talk_title',
+        'session_type',
+        'full_name',
+        'status',
+      ],
+    }
+  },
+  auto: false,
+  onSuccess(data) {
+    console.log(data)
```
harshtandiya: console.log

fossunited/api/dashboard.py
```diff
@@ -13,8 +13,18 @@ def get_event(name: str) -> dict:
...
+
+
+@frappe.whitelist(allow_guest=True)
+def get_cfp_details_from_eventname(eventname: str, fields: list) -> dict:
+    return frappe.db.get_value("FOSS Event CFP", {"event_name": eventname}, fields, as_dict=1)
+
+
+@frappe.whitelist(allow_guest=True)
+def get_proposal_list_from_eventname(eventname: str, fields: list) -> dict:
+    return frappe.get_all("FOSS Event CFP Submission", filters={"event_name": eventname}, fields=fields, page_length=99999, order_by="talk_title")
```
harshtandiya: This method can be refactored such that the details of proposal + likes on it can be transferred together. 

+ also add docstrings please

- Get all the proposals by get all
- For each proposal, get the likes count
- Stroke the likes data in another "like_data" key or something for the "proposal" object.
- Return the modified all proposals.

There's a neat function in fossunited.fossunited.utils called as `get_doc_likes`, which can be useful here.

```
from fossunited.fossunited.utils import get_doc_likes

def get_proposals_list(event_id: str, fields: list ) -> list:
    '''
    Returns the list of proposals for the given event, along with their likes data.

    Args:
        event_id (str): The event id for which the proposals are to be fetched.
        fields (list): The fields to be fetched from the proposals.

    Returns:
        list: The list of proposals for the given event.
    '''

    proposals = frappe.db.get_all(
        "FOSS Event CFP Submissions",
        {"event": event_id},
        fields=fields,
        page_length=9999,
        order_by="creation desc",
    )

    for proposal in proposals:
        proposal["likes"] = get_doc_likes("FOSS Event CFP Submission" , proposal["name"])
    
    return proposals

```

Please test/make changes as per requirements

dashboard/src/components/Breadcrumb.vue
RiyaMathew-11: target = _blank was opening breadcrumb links in new tabs - while we wanted them to be backtracked from the same tab itself - without opening a new tab. as it's usual in breadcrumbs. 

Unless that was done intentionally then ?

dashboard/src/components/proposals/ProposalBlock.vue
RiyaMathew-11: This isnt added twice - it had to be added twice to be hidden in one view and visible in another in a different place - based on the design that was provided.

dashboard/src/components/proposals/ProposalBlock.vue
RiyaMathew-11: It's not a mistake - https://github.com/fossunited/fossunited/pull/589/files/cb99fca591d632446d2fc4ebcbeb73a2b0cc31cf#r1742301231

dashboard/src/components/proposals/ProposalBlock.vue
RiyaMathew-11: will update based on this.

dashboard/src/helpers/utils.js
RiyaMathew-11: Will update the method name.

dashboard/src/pages/events/AllProposals.vue
RiyaMathew-11: This was the [first implementation](https://github.com/fossunited/fossunited/pull/589/commits/1f92ebc3e9adbe4b82b447baaa5811578033e616) - but they were blocked in Guest in view as the the APIs aren't whitelisted. 

Have mentioned this in the PR description.

fossunited/chapters/doctype/foss_chapter_event/templates/foss_chapter_event.html
RiyaMathew-11: That had come in with prettier - as the original file itself wasn't formatted it seems before. 

But as I've mentioned in the description - I have removed the Recent submissions section as planned before and added the View Submissions link as a secondary button for better visibility. Please check the screenshots for reference.

fossunited/fossunited/doctype/foss_event_cfp_submission/templates/foss_event_cfp_submission.html
RiyaMathew-11: I don't think we need a double redirect here - we could directly add that link to the dashboard? IMO a double redirect also just complicates the process. 

Especially if we are planning to move the public frontend to another path than dashboard in future.

dashboard/src/components/proposals/ProposalBlock.vue
RiyaMathew-11: Tried that first - but underline on hover on the cards seemed a bit off. This is a style choice, can consult with Jeswin and decide what's best later.

dashboard/src/pages/events/AllProposals.vue
RiyaMathew-11: Ahh they were auto created - will change them.

fossunited/api/dashboard.py
RiyaMathew-11: Originally I had used `createListResource` to fetch proposals - and `_liked_by` was not getting fetched in that. Hence I had made a separate function to fetch proposalLikes. 

However on a final test - I understood that `createListResource` won't work on public pages as it has a `get_list` invoke - and it's not whitelisted. 
So, as mentioned in the description, had to make quick whitelisted APIs that fetches the CFP and proposal details. 

I didn't append the proposalLikes fetching function here as later on there is a scope to build on top of the individual method - when implementing - likes feature on proposal cards (as I've mentioned in Additional Context). I have used the `get_doc_likes()` method here too. 

Also keeping the `proposal fetch` function separate was intentional - as that also has scope to be reused as it is - if we ever implement a feature of listing proposals submitted by a user on profile, as referred in #529, even if we don't have to bring in more extra info of likes etc there. The said `proposal fetch` API only fetches fields passed, and as no add-on strings attached to it.    

`eventname` is the current filter added but even that can be made generic later on too.

dashboard/src/pages/events/AllProposals.vue
RiyaMathew-11: https://github.com/fossunited/fossunited/pull/589#discussion_r1742364354

Please refer this to know my thoughts on why it was made that way. 

Will change to one API if needed after this too.

fossunited/api/dashboard.py
RiyaMathew-11: Let me know if it still need to be made one API

dashboard/src/components/Breadcrumb.vue
harshtandiya: It was intentional for external links actually

fossunited/fossunited/doctype/foss_event_cfp_submission/templates/foss_event_cfp_submission.html
harshtandiya: A redirect for /all page will ensure that people can share it easily. 

`/event/xyz/cfp/all` just seems natural in contrast to `/dashboard/cfp/xyz`.

Moreover, people are already used to /all now. Can't take that away from people

dashboard/src/pages/events/AllProposals.vue
RiyaMathew-11: I've used a styled version of select drop-down here. Hence it had to be monitored via the given controls. - toggles on button clicks, states etc. A normal select gives a browser rendered drop-down.

Screenshot 2024-09-03 at 10 26 33 PM

Frappe UI's select gives a browser rendered drop-down which cannot be styled as given. Hence didn't use it for this case. 
Screenshot 2024-09-03 at 10 28 30 PM


This was purely a style choice since also this is a public page. - can be changed if needed.

RiyaMathew-11: https://github.com/fossunited/fossunited/pull/589/files/cb99fca591d632446d2fc4ebcbeb73a2b0cc31cf#r1742396123

dashboard/src/pages/events/AllProposals.vue
RiyaMathew-11: https://github.com/fossunited/fossunited/pull/589/files/cb99fca591d632446d2fc4ebcbeb73a2b0cc31cf#r1742396123

fossunited/api/dashboard.py
harshtandiya: I would rather that the api endpoints are really specific rather than very generic, which may accidentally leak the info. 

For example, in this method, I don't think we would want to fetch email fields. But any signed in user can test this api and pass `email` as a field and fetch it for all proposals of an event. Ofcourse there's the requirement of knowing event id, but it's not impossible. 

On other hand, let's just fetch likes data and proposal details in a single api endpoint. Even if it isn't used somewhere else, it's data associated with that proposal.

dashboard/src/components/Breadcrumb.vue
RiyaMathew-11: Most of the external events that we add on our timeline don't take proposals in our platform, and would have their own websites.

The exception is IndiaFOSS, but we can also add All proposals link to website too.

Breadcrumbs are meant for backtrack - doesn't make sense when opened in a new tab. Had implied the same here as well - https://github.com/fossunited/fossunited/pull/572#discussion_r1734334429

fossunited/api/dashboard.py
harshtandiya: Let's remove fields as an argument that you can pass.

Let's define the fields we want to fetch in the function itself, and make sure that we are not fetching sensitive datafields.

dashboard/src/components/proposals/ProposalBlock.vue
RiyaMathew-11: https://github.com/fossunited/fossunited/pull/589/files#r1742305463

dashboard/src/pages/events/AllProposals.vue
RiyaMathew-11: https://github.com/fossunited/fossunited/pull/589/files#r1742377217

fossunited/api/dashboard.py
RiyaMathew-11: I agree that the API end points need to be specific, but even with the said implementation that you mentioned - someone can fetch emails with the API right, as fields aren't fixed defined ? 

Unless the doctype itself is protected with permissive settings ?

fossunited/api/proposal.py
RiyaMathew-11: https://github.com/fossunited/fossunited/pull/589/files#r1742364354

Please refer this to know my thoughts on why it was made that way.

Will change to one API if needed after this too.

fossunited/api/dashboard.py
RiyaMathew-11: Sure, will make the changes.

fossunited/fossunited/doctype/foss_event_cfp_submission/templates/foss_event_cfp_submission.html
RiyaMathew-11: In a way, it shouldn't be that way with redirects I feel, as we're adding more, even empty files to the codebase too. For every instance in a way, like this if we ever do a rewrite in Vue, it just gets added too - like if for events timeline or hackathon projects list. We need a better solution

If we were to establish a different frontend dir in future for public pages, would it solve the issue ?  If that's the case, would add this patch for now.

Also if it's the case of `all` pattern - would this route work - /dashboard/xyz/cfp/all ?

That's what I did [initially](https://github.com/fossunited/fossunited/pull/589/commits/bef1ec407d8f32cf3e108086067d10db1361428c), then changed to match the schedule routes

dashboard/src/pages/events/AllProposals.vue
RiyaMathew-11: Ahh! I will remove this.

fossunited/fossunited/doctype/foss_event_cfp_submission/templates/foss_event_cfp_submission.html
harshtandiya: People already are used to /all, wouldn't changing that now be too confusing without any warning whatsoever?

A simple redirect code is the easiest solution to tackle this. 

I don't think a rewrite in vue is going to happen ever, because of the complexities with 2 vue folders and routings associated with it. 

let's just add `/all` redirect, there are no downsides to it

dashboard/src/pages/events/AllProposals.vue
harshtandiya: I think you can style select fields as required. I'm pretty sure I have done the same in one of the files I mentioned above. 

I don't know if style should be kept above functionality and quality of code 😬

I really don't feel this is the way to handle this. Feels unnecessarily complex, for something that can be easier.

fossunited/api/dashboard.py
harshtandiya: There's similar code in reviewer.py, that may be helpful

dashboard/src/pages/events/AllProposals.vue
RiyaMathew-11: > I think you can style select fields as required. I'm pretty sure I have done the same in one of the files I mentioned above.

It's the drop down that I am mentioning about @harshtandiya  - you cannot style a browser rendered component consistently across to support all browsers - https://blog.logrocket.com/creating-custom-select-dropdown-css/#cross-browser-compatibility-responsiveness

And hence, as I mentioned it was custom built, which can also be used as a component in future for other drop-downs if needed.

But since you've mentioned that we don't need it, I will revert the custom component. 

> I don't know if style should be kept above functionality and quality of code 😬

I think the functionality is kept intact here, and writing of button components for select is also a used practice these days - for more control over styling and consistency. Like how the '[react-select](https://react-select.com/home)' component is written

fossunited/fossunited/doctype/foss_event_cfp_submission/templates/foss_event_cfp_submission.html
RiyaMathew-11: > A simple redirect code is the easiest solution to tackle this.

It is the easiest, but IMO not the best/optimal. And definitely not something we could follow in the longer run, as the process itself generates unused files into the codebase. There should be a better alternative.

I will add this patch on for now, due to time crunch. But will take suggestions for work-arounds.

dashboard/src/pages/events/AllProposals.vue
RiyaMathew-11: On second thoughts, I'll try using the vue-select component here rather than the custom built one - https://vue-select.org/guide/css.html#dark-mode-example

Or try styling the frappe-ui components as mentioned in.

fossunited/api/dashboard.py
RiyaMathew-11: This is updated

dashboard/src/helpers/utils.js
RiyaMathew-11: This is updated

dashboard/src/components/proposals/ProposalBlock.vue
RiyaMathew-11: updated

RiyaMathew-11:
@harshtandiya refactored workflows and some styles to remove redundancy. Please checkout and let me know if anything else to be added.

fossunited/api/proposal.py
```diff
@@ -0,0 +1,41 @@
...
+        "name",
+        "route",
+        "talk_title",
+        "session_type",
+        "full_name",
+        "status",
+    ]
+
+    proposals = frappe.get_all("FOSS Event CFP Submission", filters={"event": event}, fields=fields, page_length=99999, order_by="talk_title")
+
```
harshtandiya, line 32: nit: shouldn't we be ordering stuff by creation?

fossunited/api/proposal.py
```diff
@@ -0,0 +1,41 @@
...
+    is_cfp_anonymous = frappe.db.get_value("FOSS Event CFP", {"event": event}, "anonymise_proposals")
```
harshtandiya, line 33: why dont we do this before setting up `fields` list on line 22, and just append "full_name" based on whether the cfp is anonymous or not.

Just a thought, not mandated to apply

harshtandiya (approved):

fossunited/api/proposal.py
RiyaMathew-11: by talk proposals would be easier to find in a way - used to do it for previous years'

fossunited/api/proposal.py
RiyaMathew-11: It's the same order of time in both cases, just wanted the fetch to happen first. Can refactor it later too

PR #590, opened 2024-09-02
fix: already transferred ticket not sending emails on re-transfer

### LINKED ISSUE #588: [BUG] Issues in trasferring already transferred tickets
### Describe the bug

The conference attendees who have transferred tickets which were already transferred are not receiving the ticket pdf in their emails. 

### Expected behavior

_No response_

### To Reproduce

_No response_

### Screenshots

_No response_

### Environment

```markdown
- Device:
- OS:
- Browser:
```

### Additional context

_No response_

harshtandiya:
- [x] 🐛Bug Fix

## Description

Bug: Emails were not being sent when an already transferred ticket was being transferred again.

This happened because the notification trigger for the transfer ticket were such that, when `is_transfer_ticket` checkbox value changed from 0 to 1, then the notification was sent. 

Fixed it in this PR with `handle_already_transferred_ticket` method. The method runs when the `is_transfer_ticket` checkbox is already ticked, and it unchecks the checkbox and saves the document.

Then the other operations run as they were before.

---

Also refactored tests to use `setUp` and `tearDown` methods . And added a test to check this particular new method.

## Related Issues & Docs

closes #588

### DIFF
--- a/fossunited/ticketing/doctype/foss_event_ticket_transfer/foss_event_ticket_transfer.py
+++ b/fossunited/ticketing/doctype/foss_event_ticket_transfer/foss_event_ticket_transfer.py
@@ -53,9 +53,9 @@ class FOSSEventTicketTransfer(Document):
             )
 
     def transfer_ticket(self):
-        self.validate_ticket_exists()
         try:
             ticket = frappe.get_doc("FOSS Event Ticket", self.ticket)
+            self.handle_already_transferred_ticket(ticket)
             ticket.full_name = self.receiver_name
             ticket.email = self.receiver_email
             ticket.designation = self.designation
@@ -66,3 +66,17 @@ class FOSSEventTicketTransfer(Document):
             ticket.save(ignore_permissions=True)
         except Exception as e:
             frappe.throw(str(e), frappe.ValidationError)
+
+    def handle_already_transferred_ticket(self, ticket):
+        """
+        If the ticket that is to be transferred, is already a transferred ticket,
+        then set the `is_transfer_ticket` to 0 (unchecked).
+
+        This is done, so that any notification / doc events associated with change in `is_transfer_ticket` field
+        can be triggered again.
+        """
+        if not ticket.is_transfer_ticket:
+            return
+
+        ticket.is_transfer_ticket = 0
+        ticket.save(ignore_permissions=True)
--- a/fossunited/ticketing/doctype/foss_event_ticket_transfer/test_foss_event_ticket_transfer.py
+++ b/fossunited/ticketing/doctype/foss_event_ticket_transfer/test_foss_event_ticket_transfer.py
@@ -120,3 +120,61 @@ class TestFOSSEventTicketTransfer(FrappeTestCase):
                 }
             )
             transfer.insert()
+
+    def test_transfer_already_transferred_ticket(self):
+        fake = Faker()
+        # Given a ticket, transfer it to another user
+        sender = {
+            "full_name": fake.name(),
+            "email": fake.email(),
+        }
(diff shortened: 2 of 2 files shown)


fossunited/ticketing/doctype/foss_event_ticket_transfer/test_foss_event_ticket_transfer.py
```diff
@@ -24,12 +26,31 @@ def test_ticket_transfer(self):
         )
         event.insert()
 
+        self.event = event
+
+    def tearDown(self):
+        frappe.delete_doc("FOSS Chapter Event", self.event.name, force=True)
```
mngshm, line 32: One question, 

Why is this method deleting the document which is going to be a event ?

mngshm (approved):
The code for already transferred tickets looks legitimate and also the refactored tests.

fossunited/ticketing/doctype/foss_event_ticket_transfer/foss_event_ticket_transfer.py
```diff
@@ -66,3 +66,17 @@ def transfer_ticket(self):
...
+        then set the `is_transfer_ticket` to 0 (unchecked).
+
+        This is done, so that any notification / doc events associated with change in `is_transfer_ticket` field
+        can be triggered again.
+        """
+        if not ticket.is_transfer_ticket:
+            return
+
+        ticket.is_transfer_ticket = 0
+        ticket.save(ignore_permissions=True)
```
rahulporuri, line 82: I doubt we have the time to solve this problem in a different way but IMO the "better" way to solve this problem is to not update an existing ticket entry in the database and rely on event triggers manually but instead just create a new entry and delete the old ticket entry.

(in my understanding of the code, we're updating the same database entry, but please correct me if my understanding is wrong).

this way, we are creating a robust solution, preventing any need to toggle the transfer ticket on and off depending on whether it is a transferred ticket or not

fossunited/ticketing/doctype/foss_event_ticket_transfer/foss_event_ticket_transfer.py
```diff
@@ -53,9 +53,9 @@ def validate_status_is_pending(self):
             )
 
     def transfer_ticket(self):
-        self.validate_ticket_exists()
```
rahulporuri, line 56: not sure why we dont need to validate that the ticket exists anymore. shouldn't we still be doing it? this was preventing a valid issue that we had earlier.

fossunited/ticketing/doctype/foss_event_ticket_transfer/test_foss_event_ticket_transfer.py
harshtandiya: A good read: https://hands-on.cloud/python-unit-tests/

> Python Unit Test Teardown
 The counterpart to the setup process in unit testing is the ‘teardown’ process. The tearDown() method in the unittest module is where you clean up any resources set up. This method is automatically called after each test.

fossunited/ticketing/doctype/foss_event_ticket_transfer/foss_event_ticket_transfer.py
harshtandiya: We can design better processes in the future for sure

fossunited/ticketing/doctype/foss_event_ticket_transfer/foss_event_ticket_transfer.py
harshtandiya: It was redundant with line 37 in the same file

PR #602, opened 2024-09-12
MAINT: Add "F" flag to ruff config

rahulporuri:
- [ ] 🍕Feature
- [x] ⚙️Chore
- [ ] 🐛Bug Fix
- [ ] 🎨Style
- [ ] 🧑‍💻Refactor
- [ ] 📕Documentation
- [ ] 🏎️Optimization

## Description
This PR enables `ruff` to discover F flag issues. Ref https://docs.astral.sh/ruff/linter/#rule-selection

## Checklist
- [x] The code follows the project's coding standards.

### DIFF
--- a/fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
+++ b/fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
@@ -66,10 +66,10 @@ class FOSSChapterEvent(WebsiteGenerator):
         show_schedule: DF.Check
         show_speakers: DF.Check
         sponsor_list: DF.Table[FOSSEventSponsor]
-        status: DF.Literal["", "Draft", "Live", "Approved", "Concluded", "Cancelled"]
+        status: DF.Literal["", "Draft", "Live", "Approved", "Concluded", "Cancelled"]  # noqa: F722, F821
         t_shirt_price: DF.Currency
         ticket_form_description: DF.MarkdownEditor | None
-        tickets_status: DF.Literal["Live", "Closed"]
+        tickets_status: DF.Literal["Live", "Closed"]  # noqa: F821
         tiers: DF.Table[FOSSTicketTier]
     # end: auto-generated types
 
--- a/fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
+++ b/fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
@@ -211,6 +211,7 @@ class FOSSChapterEvent(WebsiteGenerator):
         )
         speakers = []
         for cfp in speaker_cfps:
+            user = frappe.get_doc(USER_PROFILE, {"email": cfp.submitted_by})
             speakers.append(
                 {
                     "full_name": cfp.full_name,
--- a/fossunited/foss_hackathon/doctype/foss_hackathon_team/foss_hackathon_team.py
+++ b/fossunited/foss_hackathon/doctype/foss_hackathon_team/foss_hackathon_team.py
@@ -1,7 +1,6 @@
 # Copyright (c) 2024, Frappe x FOSSUnited and contributors
 # For license information, please see license.txt
 
-import frappe
 from frappe.model.document import Document
 
 
--- a/fossunited/foss_hackathon/doctype/foss_hackathon_team/foss_hackathon_team.py
+++ b/fossunited/foss_hackathon/doctype/foss_hackathon_team/foss_hackathon_team.py
@@ -22,7 +22,7 @@ class FOSSHackathonTeam(Document):
         looking_for_members: DF.Check
         members: DF.Table[FOSSHackathonTeamMember]
         partner_project: DF.Link | None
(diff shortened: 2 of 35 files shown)


fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
```diff
@@ -211,6 +211,7 @@ def get_speakers(self):
         )
         speakers = []
         for cfp in speaker_cfps:
+            user = frappe.get_doc(USER_PROFILE, {"email": cfp.submitted_by})
```
rahulporuri, line 214: see line 218 - `user.profile_photo` is used but `user` isn't defined. this is the one real issue that was caught by introducing the `F` flag

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
```diff
@@ -112,7 +112,7 @@ def get_context(self, context):
             "Administrator",
             self.user,
         ):
-            frappe.throw(_("Profile Not Found"), frappe.DoesNotExistError)
```
rahulporuri, line 115: it's not clear to me what `_` is supposed to mean here or what it is supposed to do.

fossunited/integrations/github.py
```diff
@@ -13,7 +13,7 @@ def __init__(self):
             access_token = github_settings.get_password("access_token")
             self.github = Github(access_token)
         except Exception as e:
-            frappe.log_error(frappe.get_traceback(), _("Github Settings not found"))
```
rahulporuri, line 17: it's not clear to me what `_` is supposed to mean here or what it is supposed to do.

RiyaMathew-11: If I am not wrong, it's used in translation mechanism - adding `_()` makes the text translatable to other languages, if there are browser workflows that support it. 

Otherwise, it brings in the original string itself

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
RiyaMathew-11: https://github.com/fossunited/fossunited/pull/602#discussion_r1757129230

fossunited/integrations/github.py
rahulporuri: thanks for the clarification. i'm happy to add it back but I don't see it's use in our context

fossunited/foss_hackathon/doctype/foss_hackathon_team/foss_hackathon_team.py
```diff
@@ -1,7 +1,6 @@
 # Copyright (c) 2024, Frappe x FOSSUnited and contributors
 # For license information, please see license.txt
 
-import frappe
```
RiyaMathew-11, line 4: I am not sure if `import frappe` should be something that can be safely removed - would be worth a check.

fossunited/integrations/github.py
RiyaMathew-11: @harshtandiya can comment on this better, as to why it's used here specifically. 

But in general sense, it's something that improves accessibility as well.

fossunited/foss_hackathon/doctype/foss_hackathon_team/foss_hackathon_team.py
rahulporuri: I'd be very surprised if the import itself has side-effects. The `frappe` import itself isn't being used in the codebase, which ruff can attest to.

fossunited/integrations/github.py
harshtandiya: It works either way imo

harshtandiya (approved):

fossunited/foss_hackathon/doctype/foss_hackathon_team/foss_hackathon_team.py
RiyaMathew-11: Yep, it's just a doubt/question - if the file itself imports frappe first and then the subsequent modules - frappe.types or frappe.model.....

Just wanted to understand the reason why are they autogenerated in the first place.

rahulporuri: > Just wanted to understand the reason why are they autogenerated in the first place.

From what I've seen with other frameworks, it reduces one additional thing that a developer has to add manually e.g. just using the import or updating it as necessary.

An import statement having a side-effect is a big no-no in Python and frowned upon.

fossunited/integrations/github.py
rahulporuri: thank you. not reverting the change and merging as-is. I've also resolved the other comment regarding the same

rahulporuri:
@RiyaMathew-11 @harshtandiya i used squash merge strategy to merge this PR - only one commit is created on the develop branch instead of the 8 commits on this branch being added to develop + a merge commit. I find that the timeline becomes cleaner/easier to understand and easier go back at PRs from the commits timeline

PR #617, opened 2024-09-16
chore: selective update permissions in razorpay payments

RiyaMathew-11:
- [x] ⚙️Chore
- [x] 🧑‍💻Refactor

## Description

The PR adds selected permissions levels on Payments' doctypes to facilitate edits by system managers and admins. The original implementation only had read-only permissions. 

Permission changes: 
- Removed read-only status of all customer-section fields. 
- Added perm level 1 for the customer section fields. 
- Perm level means - read and write permissions to only admin and system managers. 
- Permission level 1 granted for system manager and admin.

With these changes, admin and system managers would be able to edit these fields safely via desk.

### Test cases added: 

- Validate admin permissions of creating an invoice. 
- Validate that a normal user cannot create an invoice.

## Checklist
- [x] ~I have updated the documentation (If applicable)~.
- [x] ~The code follows the project's coding standards~.


fossunited/doctype_ids.py
```diff
@@ -33,3 +33,9 @@
 # Ticket-related identifiers
 EVENT_TICKET = "FOSS Event Ticket"
 TICKET_TRANSFER = "FOSS Event Ticket Transfer"
+
+# Payments
+
+RAZORPAY_PAYMENT = "Razorpay Payment"
+RAZORPAY_SETTINGS = "Razorpay Settings"
+RAZORPAY_WEBHOOK_LOGS = "Razorpay Webhook Log"
```
rahulporuri: minor suggestion - as you're adding the constants, could you also update the relevant usage in the codebase to use these constants?

fossunited/payments/doctype/razorpay_payment/razorpay_payment.json
```diff
@@ -137,20 +137,21 @@
       "fieldname": "gstn",
       "fieldtype": "Data",
       "label": "GSTN",
-      "read_only": 1
+      "permlevel": 1,
+   "read_only": 1
```
rahulporuri: nit - weird formatting, might be a bug in the autogeneration

```suggestion
      "permlevel": 1,
      "read_only": 1
```

fossunited/payments/doctype/razorpay_payment/razorpay_payment.json
```diff
@@ -189,7 +190,40 @@
...
+   "email": 1,
+   "export": 1,
+   "permlevel": 1,
+   "print": 1,
+   "read": 1,
+   "report": 1,
+   "role": "Administrator",
+   "share": 1,
+   "write": 1
+  }
```
rahulporuri: potentially noob question - why are there two entries for `Administrator`?

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
rahulporuri, line 1: :heart:

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
```diff
@@ -0,0 +1,137 @@
...
+class TestRazorpayPayment(FrappeTestCase):
+    def setUp(self):
+        self.admin_user = frappe.get_doc("User", "Administrator")
+        self.normal_user = frappe.get_doc("User", "test1@example.com")
+
+    def tearDown(self):
+        frappe.set_user("Administrator")
+
+        # Delete any created Razorpay Payment documents
+        payments = frappe.get_all(RAZORPAY_PAYMENT, filters={"document_name": self.event})
```
rahulporuri, line 22: nit - `self.event` won't exist unless the `create_payment_document` is called. `unittest.TestCase.tearDown` methods are expected to never fail because they are expected to handle any cleanup because of test failures.

because the tests need a payment invoice, you could update the `setUp` to run the relevant method, ensuring that `self.event` always exists on the instance when `tearDown` is executed.

alternatively, you could set it to `None` and only run this code if `self.event is not None`

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
```diff
@@ -0,0 +1,137 @@
...
+    def generate_random_permalink(max_length=20):
+        words = fake.words(nb=3)
+        permalink = "_".join(word.lower() for word in words)
+
+        return permalink[:max_length]
+
+    def create_dummy_event(self):
+        event = frappe.get_doc(
+            {
+                "doctype": "FOSS Chapter Event",
```
rahulporuri: nit (more than one usage of `"FOSS Chapter Event"` in this file) - this should not be available via `fossunited.doctype_ids`

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
```diff
@@ -0,0 +1,137 @@
...
+                "enabled": tier.enabled,
+                "idx": tier.idx,
+                "maximum_tickets": tier.maximum_tickets,
+                "modified": str(tier.modified),
+                "modified_by": tier.modified_by,
+                "name": tier.name,
+                "owner": tier.owner,
+                "parent": tier.parent,
+                "parentfield": "tiers",
+                "parenttype": "FOSS Chapter Event",
```
rahulporuri: nit - available via `fossunited.doctype_ids`

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
```diff
@@ -0,0 +1,137 @@
...
+                "price": tier.price,
+                "title": tier.title,
+                "valid_till": str(tier.valid_till),
+            },
+        }
+
+        invoice = frappe.get_doc(
+            {
+                "doctype": "Razorpay Payment",
```
rahulporuri: nit - use the constant defined in this PR

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
```diff
@@ -0,0 +1,137 @@
...
+            frappe.db.commit()
+            return invoice
+        except frappe.PermissionError:
+            frappe.db.rollback()
+            return None
+
+    def test_admin_can_create_invoice(self):
+        invoice = self.create_payment_document(self.admin_user)
+        self.assertIsNotNone(invoice)
+        if invoice:
```
rahulporuri: nit - `if invoice` will always be truthy because we only execute this statement if `self.assertIsNotNone(invoice)` passes - i don't think we need the conditional here

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
```diff
@@ -0,0 +1,137 @@
...
+            self.assertEqual(invoice.doctype, RAZORPAY_PAYMENT)
+            self.assertEqual(invoice.owner, self.admin_user.name)
+
+    def test_normal_user_cannot_create_invoice(self):
+        invoice = self.create_payment_document(self.normal_user)
+        self.assertIsNone(invoice)
```
rahulporuri: non-actionable comment - another way to write the test would be to explicitly catch the `frappe.PermissionError`. the system will raise a `PermissionError` if this situation somehow happens so the test should emulate the system behavior.

```suggestion
        with self.assertRaises(frappe.PermissionError):
            invoice = self.create_payment_document(self.normal_user)
```

+ the necessary change in `self.create_payment_document` that re-raises the exception after performing the rollback

rahulporuri (approved):
Overall LGTM, apart from a bunch of minor nitpicky and one non-actionable comment.

@harshtandiya might be have another pair of eyes on the permissions bit

harshtandiya:
Can you please explain the perm level decisions? It is very unclear by just looking at the code

fossunited/payments/doctype/razorpay_payment/razorpay_payment.json
harshtandiya: 💯

harshtandiya: I think the formatting is a bit off for every json file. Can you check @RiyaMathew-11 ? Make sure you run prettier with the latest settings of repo

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
```diff
@@ -0,0 +1,137 @@
...
+        frappe.db.commit()
+
+    @staticmethod
+    def generate_random_permalink(max_length=20):
+        words = fake.words(nb=3)
+        permalink = "_".join(word.lower() for word in words)
+
+        return permalink[:max_length]
+
+    def create_dummy_event(self):
```
harshtandiya, line 31: Why not just create event in `setUp` and assign it as `self.event`? The `tearDown` will make more sense in that way(reference to above comment by Rahul)

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
```diff
@@ -0,0 +1,137 @@
...
+
+        # Delete created event
+        frappe.delete_doc(EVENT, self.event)
+
+        frappe.db.commit()
+
+    @staticmethod
+    def generate_random_permalink(max_length=20):
+        words = fake.words(nb=3)
+        permalink = "_".join(word.lower() for word in words)
```
harshtandiya: this can be one line when you're creating an event. Having a whole method for this seems like an overkill

```py
 "event_permalink": fake.slug().replace("-", "_"),
```

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
```diff
@@ -0,0 +1,137 @@
...
+
+        return permalink[:max_length]
+
+    def create_dummy_event(self):
+        event = frappe.get_doc(
+            {
+                "doctype": "FOSS Chapter Event",
+                "event_name": fake.catch_phrase(),
+                "event_permalink": self.generate_random_permalink(),
```
harshtandiya: Again, just one line would suffice imo

```py
"event_permalink": fake.slug().replace("-", "_"),
```

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
```diff
@@ -0,0 +1,137 @@
...
+                "status": "Live",
+                "event_type": "CityFOSS Conference",
+                "event_start_date": "2024-09-05 10:00:00",
+                "event_end_date": "2024-09-05 13:00:00",
```
harshtandiya, line 40: nit: Why not use fake.date() or datetime.today()?

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
```diff
@@ -0,0 +1,137 @@
...
+                "event_description": fake.paragraph(),
+                "t_shirt_price": 100,  # Set t-shirt price
+                "is_paid_event": 1,
+                "tiers": [{"title": "General", "price": 200}],
+            }
+        ).insert(ignore_permissions=True)
+        return event.name
```
harshtandiya, line 47: again, having this in setup itself would be nicer.

Take reference: https://github.com/fossunited/fossunited/blob/develop/fossunited/ticketing/doctype/foss_event_ticket/test_foss_event_ticket.py

harshtandiya:
I was of the idea that there were tests added for the method that works along with the frontend: https://github.com/fossunited/fossunited/blob/develop/fossunited/api/dashboard.py#L26

Maybe we can add that test too later

fossunited/payments/doctype/razorpay_payment/razorpay_payment.json
RiyaMathew-11: Permission levels 0 and 1 - for each case is listed separately

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
RiyaMathew-11: I've added self.event in setup to make the flow proper. It seems when did a merge commit to resolve conflicts, it got reverted in a way. 

Fixing this in the next commit

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
RiyaMathew-11: Added as `EVENT` via `doctype_id` imports

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
RiyaMathew-11: Ah - this was a misstep, changed it

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
RiyaMathew-11: Makes sense - changing it.

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
RiyaMathew-11: Added

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
RiyaMathew-11: I kind of wanted to reuse this method later for test-cases I planned to add for events' module. As permalink creation itself has a lot of validations.

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
RiyaMathew-11: Didn't use datetime to reduce imports. Plus I've kind of added hard-coded values now for those fields that won't create much implication if the value is static. 

But can make it as faker value.

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
RiyaMathew-11: https://github.com/fossunited/fossunited/pull/617#discussion_r1768387463

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
RiyaMathew-11: I will change it to the fake value

fossunited/payments/doctype/razorpay_payment/razorpay_payment.json
RiyaMathew-11: I've run prettier on the file again. Looks ok now

fossunited/doctype_ids.py
RiyaMathew-11: Sure

RiyaMathew-11:
> Can you please explain the perm level decisions? It is very unclear by just looking at the code

Added this in the PR description

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
```diff
@@ -0,0 +1,136 @@
...
+            }
+        )
+
+        try:
+            invoice.insert()
+            frappe.db.commit()
+            return invoice
+        except frappe.PermissionError:
+            frappe.db.rollback()
+            return None
```
harshtandiya: CI is failing because of this block. You need to throw the error rather than returning `None`

RiyaMathew-11:
@rahulporuri @harshtandiya the merge errors, formatting errors have been cleared. I've added changes based on a couple of suggestions too. 

Do give a final check and let me know if this is good to go ?

rahulporuri (approved):
Still LGTM, from what i can see

harshtandiya:
Just test ticket buying as a guest user, and then we are good to merge

harshtandiya (approved):

RiyaMathew-11:
> Just test ticket buying as a guest user, and then we are good to merge

Works as intended: 
Screenshot 2024-09-25 at 6 11 05 PM
Screenshot 2024-09-25 at 6 11 47 PM

RiyaMathew-11:
Merging this PR based on the tests @harshtandiya @rahulporuri

Issue #424, opened 2024-07-02
[FEATURE] Implementing proper routes for user and chapters on the platform


mngshm:
### Feature description

There is a lot of confusion between the routes of a user profile and a chapter profile.

harshtandiya:
Let's have users under `/u/`, ex: `fossunited.org/u/harsh` ?

For cities, we can either have them as they are currently, i.e, `fossunited.org/<city_name>` or we can move them under `/c/`. Thoughts on this?

rahulporuri:
- `/u/` or `/user/` for individual users
- `/c/` or `/community/` for communities, city chapters & student clubs e.g. `/community/hyderabad` and `/community/iiit-kalyani`. i don't think the URLs need to differentiate between them

harshtandiya:
I think having the cities at the top level without `/c/` would work too. Cities are limited anyway. For student clubs, we should indeed move them under `/c/` imo.

After the refactor of chapter routes, the event routes for City Chapters can be made to be like this:
`fossunited.org/bengaluru/2024/june`

or if we go with `/c/` for city communities too:
`fossunited.org/c/bengaluru/2024/june`


For clubs it can be under `/events/` or we can change it to be under the chapter's path:
`fossunited.org/c/iiit-kalyani/linux-installation-2024/`
`fossunited.org/c/iiit-kalyani/2024/catch-the-flag`

I need help to have more clarity on this.


Pros for having cities on top level:
- Easier to access chapter pages
- Shorter routes for events

Cons:
- More code to handle different cases but manageable

rahulporuri:
> I think having the cities at the top level without /c/ would work too. Cities are limited anyway. For student clubs, we should indeed move them under /c/ IMO.

I would prefer strictly namespacing the cities now, while we're making the changes. Cities are limited but we could have 100s of city chapter a decade from now, in different stages of activity

> After the refactor of chapter routes, the event routes for City Chapters can be made to be like this:

for the events, I would recommend the format `/c/<chapter-identifier>/year/<event-identifier>` - the chapter identifiers have to be unique i.e. a city chapter will use the city name and clubs in the city will use the college name. the event identifiers have to also be unique e.g. month name if only one event happens per month or unique identifiers like mini hackathon, cityfoss, x workshop, etc.

@captn3m0 thoughts?

captn3m0:
[...]

PR #644, opened 2024-09-28
REFACTOR!: chapter & event routes

harshtandiya:
- [x] ⚙️Chore
- [x] 🧑‍💻Refactor

## Description

See: #332 & #424 

There have been discussions of refactoring the chapter routes to come under `/c/`.
This means that what previously would be something like `fossunited.org/hyderabad` would change to become `fossunited.org/c/hyderabad`. 
Similarly, refactoring the events routes from `/event` to come under the chapter's route made more sense. 
Currently, all events are named in such a way: `fossunited.org/event/hyd-sep-2024`. But this PR changes that to be something of this pattern: `fossunited.org/c/hyderabad/2024/sep`

These are the formats followed:
- For Chapter: `/c/<chapter_slug>`
- For Event: `/c/<chapter_slug>/<event_permalink>`

---

Additionally:
- Removed the validations which were stopping organizers from adding `/` in permalinks. This will enable a greater sense of freedom in cause of event routes.
- Added a couple of tests for the new routes. Which ensures that no two events in a chapter have same kind of routes
- Routing was breaking for CFP All & Schedule dashboard pages, this PR fixes that.
- CFP & RSVP submission page was breaking, fixed that too.

## Related Issues & Docs
#424 

image

### DIFF
--- a/dashboard/src/components/CopyToClipboardButton.vue
+++ b/dashboard/src/components/CopyToClipboardButton.vue
@@ -1,0 +1,27 @@
+<template>
+  <Tooltip :text="tooltipText">
+    <Button icon="copy" class="w-4" variant="ghost" @click="handleCopy" />
+  </Tooltip>
+</template>
+<script setup>
+import { Tooltip } from 'frappe-ui'
+import { copyToClipboard } from '@/helpers/utils'
+import { ref } from 'vue'
+
+const tooltipText = ref('Copy to clipboard')
+
+const props = defineProps({
+  value: {
+    type: String,
+    required: true,
+  },
+})
+
+const handleCopy = () => {
+  copyToClipboard(props.value)
+  tooltipText.value = 'Copied!'
+  setTimeout(() => {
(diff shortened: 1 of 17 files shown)


rahulporuri:
@harshtandiya CI seems to be failing. can you take a look? I will do a shallow review in the meanwhile

rahulporuri:
As discussed in the daily tech standup, we want to implement redirects for all of the existing URLs to ensure that the existing URLs don't break - but we will also notify the community that the redirect will only work temporarily (e.g. one month or until end of year) and the old URLs will not work after the duration.

harshtandiya:
@rahulporuri 

We should add redirects for the live events. That can only be done after this update is merged and deployed. 

For all the other past events, I think we should just go ahead and change the routes for them. Past events can be found from the event timeline and chapter pages anyways, so this should be good to go ahead with

rahulporuri (approved):

PR #657, opened 2024-10-09
refactor(dashboard): event creation page

### LINKED ISSUE #656: Better error messages for event creation
### Issue description

Many chapters are unable to create events, simply because they are doing some error or leaving out some form field empty during the event creation.

UX can be made better by handling errors in a better way and being upfront about the mandatory fields.

harshtandiya:
- Refactored the event creation page in dashboard to do better error handling.
- Mandatory fields are marked as such, and an error is thrown on submit if the field is left empty by the user.
- Minor style changes.
- Show the event link as per the permalink being set.

- [x] ⚙️Chore
- [x] 🧑‍💻Refactor

## Related Issues & Docs

fixes partially #364 
closes #656 

image

### DIFF
--- a/dashboard/src/pages/CreateEventForm.vue
+++ b/dashboard/src/pages/CreateEventForm.vue
@@ -4,84 +4,120 @@
       <ChapterHeader :chapter="chapter" />
     </div>
     <div class="flex flex-col md:flex-row gap-2 justify-between mt-6 pb-2 border-b">
-      <div class="text-xl font-semibold">Create Event</div>
-      <Button label="Create" :variant="'solid'" size="md" @click="createEvent" />
+      <div class="prose">
+        <h2>Create Event</h2>
+      </div>
     </div>
     <div v-if="eventTypeOptions.data">
-      <div class="flex flex-col gap-3 my-6">
-        <div class="font-semibold text-gray-800 border-b-2 pb-2">Details</div>
-        <div class="p-2 my-1 grid sm:grid-cols-1 md:grid-cols-2 gap-x-4 gap-y-6">
-          <FormControl
-            v-model="temp_event.event_name"
-            :type="'text'"
-            size="md"
-            label="Event Name"
-          />
-          <FormControl
-            v-model="temp_event.event_permalink"
-            :type="'text'"
-            size="md"
-            label="Event Permalink"
-            description="Only enter the event endpoint. Events will be rendered at event/<event_permalink>"
-          />
-          <FormControl
-            v-model="temp_event.status"
-            :type="'select'"
-            :options="[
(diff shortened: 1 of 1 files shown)


dashboard/src/pages/CreateEventForm.vue
```diff
@@ -4,84 +4,120 @@
...
+                {
+                  label: 'Draft',
+                  value: 'Draft',
+                },
+                {
+                  label: 'Live',
+                  value: 'Live',
+                },
+                {
+                  label: 'Concluded',
```
RiyaMathew-11: I think we needn't include Concluded and Cancelled in a create event form. They are CTAs that come after an event is made. Draft and Live makes sense

dashboard/src/pages/CreateEventForm.vue
```diff
@@ -4,84 +4,120 @@
...
+          </div>
+          <div class="grid sm:grid-cols-1 md:grid-cols-2 gap-x-4 gap-y-6">
+            <FormControl
+              v-model="temp_event.event_start_date"
+              :type="'datetime-local'"
+              size="md"
+              label="Event Start Date&ast;"
+            />
+            <FormControl
+              v-model="temp_event.event_end_date"
```
RiyaMathew-11, line 96: Here would be good to add a validation to ensure that end_date > start_date always

dashboard/src/pages/CreateEventForm.vue
```diff
@@ -4,84 +4,120 @@
...
-            label="Event Type"
-          />
-          <TextEditor
-            label="Event Description"
-            class="col-span-2"
-            placeholder="Write an event description"
-            :model-value="temp_event.event_description"
-            @update:model-value="($event) => (temp_event.event_description = $event)"
-          />
+      <div class="flex flex-col gap-5 my-6">
```
RiyaMathew-11, line 12: I don't see it in the UI - but I think it would also be better to add a red asterisk or so - to mark fields that are important. 

That way the occurrence of MandatoryFields Error is actually minimised.

RiyaMathew-11 (approved):
Added a few comments - functionally I think the PR LGTM!

dashboard/src/pages/CreateEventForm.vue
harshtandiya: makes sense, changing

dashboard/src/pages/CreateEventForm.vue
harshtandiya: I think just an asterisk would suffice, as other forms too has an asterisk with the label (gray not red). Gray asterisk is there in this form as well

PR #670, opened 2024-10-21
FEAT: Emailing function for organizers

### LINKED ISSUE #702: [FEATURE] Custom Emailing for Events
### Feature description

- Create dashboard for organizers which will enable them to create custom email campaigns and send it to the participants/cfp proposers

harshtandiya:

- [x] 🍕Feature

## Description

This PR adds the `Emailing` function per event basis.

A dashboard page is created for chapter organizers to use this feature.

Functionalities:
- Create Email Campaigns linked to specific chapter & event
- Send it to pre-made email groups such as: `Event Participants`, `CFP Proposers` `Accepted` and `Rejected` CFP proposers etc.
- Add attachments to campaigns.
- Schedule the sending of a campaign.
- Send test emails
- Send bulk emails

Additionally done in this PR:
- created a `Drawer` ui component similar to [tailwind's example](https://tailwindui.com/components/application-ui/overlays/drawers)
- added `lodash` package to help with Object related operations, such as deepCopy, equality check etc.
- bumped frappe-ui from 0.1.50 to 0.1.86
- bumped tailwind to `3.2.7`  

## Related Issues & Docs

partial #107 
closes #702

#### Creating a Campaign
[Screencast from 2024-11-29 17-21-10.webm](https://github.com/user-attachments/assets/4fa96415-efc5-4750-9957-dd8fcad9d419)

#### Modifying a campaign
[Screencast from 2024-11-29 17-22-40.webm](https://github.com/user-attachments/assets/e368cd41-0f34-4256-91d2-f1903bfd12f9)

#### Schedule a campaign
[Screencast from 2024-11-29 17-23-43.webm](https://github.com/user-attachments/assets/beaddc46-770a-4cc6-937b-95713eccee5c)

#### Send Test Email
[Screencast from 2024-11-29 17-24-20.webm](https://github.com/user-attachments/assets/8900e12c-8355-45b1-88ab-749a7a561037)

#### Send Email
[Screencast from 2024-11-29 17-27-23.webm](https://github.com/user-attachments/assets/11eb8f61-1ccc-4c5d-a934-4c6fd116178a)


rahulporuri:
@harshtandiya the branch has conflicts. please resolve them.

harshtandiya:
@rahulporuri done

rahulporuri:
@harshtandiya CI is failing now. Looks like a genuine issue. I'll go ahead and start reviewing the PR in the meanwhile

```
frappe.exceptions.LinkExistsError: Cannot delete or cancel because FOSS Chapter Event <a href="/app/Form/FOSS Chapter Event/mgdg98iaq3">mgdg98iaq3</a> is linked with Email Group <a href="/app/Form/Email Group/Rejected Proposers-Streamlined composite Local Area Network-None-mgdg98iaq3">Rejected Proposers-Streamlined composite Local Area Network-None-mgdg98iaq3</a> 
```

dashboard/src/components/mailing/CampaignList.vue
```diff
@@ -0,0 +1,71 @@
+<!-- eslint-disable vue/no-unused-vars -->
```
rahulporuri: nitpicky question : can i ask why the eslint-disable?

dashboard/src/components/ui/CodeEditor.vue
```diff
@@ -0,0 +1,56 @@
...
+
+const getCodemirrorStates = () => {
+  const state = view.value.state
+  const ranges = state.selection.ranges
+  const selected = ranges.reduce((r, range) => r + range.to - range.from, 0)
+  const cursor = ranges[0].anchor
+  const length = state.doc.length
+  const lines = state.doc.lines
+}
+</script>
```
rahulporuri: q: not sure why we need a codeeditor.

is it to support the html email type? if so, i would prefer not being able to write emails in HTML and reducing this code instead of adding the additional dependencies and code

dashboard/src/helpers/emailing.js
```diff
@@ -0,0 +1,27 @@
...
+
+  if (data.content_type == 'Rich Text' && !data.message.length) {
+    errors.push('Message is required.')
+  }
+  if (data.content_type == 'Markdown' && !data.message_md.length) {
+    errors.push('Message is required.')
+  }
+  if (data.content_type == 'HTML' && !data.message_html.length) {
+    errors.push('Message is required.')
+  }
```
rahulporuri: do we really need to support all of these flavors? for now, can't we just settle for `Rich Text` as the only option?

fossunited/api/emailing.py
```diff
@@ -0,0 +1,451 @@
+from typing import Literal
+
+import frappe
+
+from fossunited.api.chapter import check_if_chapter_member
+from fossunited.doctype_ids import EVENT
+
+EMAIL_GROUP_TYPES = Literal[
+    "Chapter Master",
```
rahulporuri: nit - I would prefer not using the Master/Slave terminology. Can we go with `Chapter main`?

And does `Chapter main` mean all people who have ever attended a chapter meetup?

fossunited/api/emailing.py
```diff
@@ -0,0 +1,451 @@
...
+    Create an email group linked to the event
+
+    Args:
+        event: event id
+        session_user: session user id
+    """
+    _event = frappe.get_doc(EVENT, event_id)
+
+    # if not check_if_chapter_member(_event.chapter, session_user):
+    #     raise frappe.PermissionError("You are not authorised for this action.")
```
rahulporuri: let's not check-in commented out/unused code

fossunited/api/emailing.py
```diff
@@ -0,0 +1,451 @@
...
+
+    if frappe.db.exists(
+        "Email Group", {"event": event_id, "chapter": _event.chapter, "group_type": type}
+    ):
+        raise frappe.ValidationError("Email Group already exists for this event")
+
+    group_title = f"{type}-{_event.event_name}-{_event.chapter_name}-{event_id}"
+    _group_title = group_title[:140]
```
rahulporuri: this feels suboptimal. i guess what we're trying to do here is to create a unique title for the doc. and the `title` field is used elsewhere, which is why it needs to be meaningful, is that right?

If so, can we just remove the `{type}` and `{_event.chapter_name}` from the format because they are already stored on the `group` object. we can reconstruct the string from the data if necessary.

Can you demo a few examples here i.e. what does `group_title` look like?

fossunited/api/emailing.py
```diff
@@ -0,0 +1,451 @@
...
+        chapter: chapter id
+    """
+    _event = event
+    _chapter = chapter
+
+    if not _event and not _chapter:
+        frappe.throw("Atleast one of event or chapter is required")
+
+    if not _chapter:
+        _chapter = frappe.db.get_value("FOSS Chapter Event", _event, ["chapter"])
```
rahulporuri, line 95: nit - shouldn't the UI always have information about the event and chapter? Under what conditions do we need to worry about this code?

i would prefer removing these conditionals unless we absolutely need them otherwise we need to test these conditions to ensure that the exceptions are actually raised, especially because this is a public-facing API

fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
```diff
@@ -99,6 +104,15 @@ def before_save(self):
             self.update_published_status()
         self.set_route()
 
+    def create_email_groups(self):
+        for group in [
+            "Event Participants",
+            "CFP Proposers",
+            "Accepted Proposers",
+            "Rejected Proposers",
```
rahulporuri, line 115: The list of options here differs from the earlier list of email groups. If this is intentional, I'm not sure why it is so.

fossunited/overrides/newsletter_master.py
rahulporuri, line 1: nit - same comment as earlier against the use of master/slave terminology. let's instead use things like `main` or something along those lines

fossunited/hooks.py
```diff
@@ -84,3 +84,15 @@
...
+# ----------------
+after_install = "fossunited.install.after_install"
+
+# Uninstallation
+# ----------------
+before_uninstall = "fossunited.uninstall.before_uninstall"
+
+override_doctype_class = {
+    "Newsletter": "fossunited.overrides.newsletter_master.NewsletterMaster",
+}
```
rahulporuri: q: not sure why we need a different `Newsletter` doctype.

fossunited/hooks.py
```diff
@@ -84,3 +84,15 @@
...
     {"source": r"c/(.+)/schedule", "target": r"/dashboard/schedule/\1"},
 ]
+
+# Installation
+# ----------------
+after_install = "fossunited.install.after_install"
+
+# Uninstallation
+# ----------------
+before_uninstall = "fossunited.uninstall.before_uninstall"
```
rahulporuri, line 94: q. not sure why we need the new after install/before uninstall stuff either

fossunited/setup.py
```diff
@@ -0,0 +1,109 @@
...
+                "dt": doctype,
+            },
+        )
+
+        frappe.clear_cache(doctype=doctype)
+
+
+def get_custom_fields():
+    """
+    FOSSUnited specific custom fields that need to be added to base Frappe doctypes
```
rahulporuri, line 31: q. why do these custom fields need to be added to the base doctypes? we didn't have to do this earlier so why now?

dashboard/src/components/ui/CodeEditor.vue
harshtandiya: HTML and Markdown both are handled through this. I think offering different message input types would be a really good UX, and i can already see people asking for that feature sometime in future even if we remove it rn

fossunited/setup.py
harshtandiya: Newsletter is  a core frappe doctype. We are customizing this core doctype at the installation of our app, hence the `before_install` and `after_install` hooks earlier. 

We are doing this, because there are a ton of functionalities included already in the core [Newsletter doctype,](https://github.com/frappe/frappe/blob/develop/frappe/email/doctype/newsletter/newsletter.py) so rather than rewriting that code, the more efficient way is to add these 2 custom fields and use the newsletter doctype as we want.

We never had to do this earlier because there was no use case for it before.

fossunited/hooks.py
harshtandiya: for adding the custom fields of `event` & `chapter` to core newsletter doctype. This is a more standard way to customizing core doctypes

see example: [Frappe HR](https://github.com/frappe/hrms/blob/develop/hrms/hooks.py)

fossunited/hooks.py
harshtandiya: The way newsletter naming works in the frappe framework, is that the name = newsletter subject. This will definitely cause problems in our usecase. Here, we are overriding the newsletter doctype and adding a [`autoname`](https://frappeframework.com/docs/user/en/basics/doctypes/controllers) controller to have the name to be a random hash

fossunited/api/emailing.py
harshtandiya: 1. will change
2. Yes

fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
harshtandiya: We should not be needing to create Chapter Master and Other group at this moment. The first should happen at chapter create and Other option should be used in future when introducing custom group function

fossunited/overrides/newsletter_master.py
```diff
@@ -0,0 +1,16 @@
+import frappe
+from frappe.email.doctype.newsletter.newsletter import Newsletter
+from frappe.utils import random_string
+
+
+class NewsletterMaster(Newsletter):
```
rahulporuri: comment from review call - we need to document the reason why we need to subclass and override the `autoname` method here. it isn't obvious why we need to do so from the code itself. in my undestanding, we need to do so because the subject of the email is the name, potentially causing `name` conflicts in the database. right?

fossunited/setup.py
rahulporuri: makes sense - but it will be nice if we could document the rationale for why we need these fields in the underlying doctypes.

fossunited/api/emailing.py
rahulporuri: based on review call, the ability to send email/campaign from the chapter and the event is a specific feature request - because of this, an event email might not have chapter info and vice-versa for the chapter email.

things still feel suboptimal so we should likely rearchitect the system to store and pass information from the frontend where relevant and do fewer database lookups in the backend if possible. we should break this out into a separate issue.

rahulporuri (commented):
Mostly good to go. A few review comments based on review call.

fossunited/api/emailing.py
harshtandiya: image

maybe the best way to handle this would be to extend email group to have a custom auto naming similar to Newsletter doctype?

harshtandiya: That way we can have the IDs to be random hash

rahulporuri: > maybe the best way to handle this would be to extend email group to have a custom auto naming similar to Newsletter doctype?

i'm not sure.

also, what's wrong with `{type}-{event id}` like I had mentioned earlier? The `FOSS Meetup-BANGALORE` doesn't add any meaningful information to the id and even if there is a condition for the ID to be unique, because the event ID is unique, this should be fine. Right?

rahulporuri (approved):
LGTM barring the unresolved comment regd ID convention - it isnt a blocker but it's good to resolve.

fossunited/api/emailing.py
harshtandiya: this would work too, i was suggesting an alternative. Will be making changes as per suggestions

PR #678, opened 2024-10-30
refactor: move profiles under `/u/` subroute

### LINKED ISSUE #424: [FEATURE] Implementing proper routes for user and chapters on the platform

harshtandiya:

- [x] ⚙️Chore
- [x] 🐛Bug Fix
- [x] 🧑‍💻Refactor

## Description

- Changed the logic for `update_profile` method in `fossunited.api.profiles`. New approach uses the document api `get_doc` to make changes to the profile details. This enables us to use the document controllers such as `before_save`, `on_update` etc, to be run every time the profile is updated. This helps in robust validations.

- Minor fix: the path for default user cover images was incorrect.

- Major: refactored the `before_save` controller to change the user routes to prefix with `u/` followed by `<username>`

#### For existing users:
There will be a need to do a mass update of routes for all the existing users. Since this route rule is written in the `before_save` controller, a simple `unpublish -> publish` should rewrite the route for a user. 

## Related Issues & Docs
Closes #424

### DIFF
--- a/fossunited/api/profile.py
+++ b/fossunited/api/profile.py
@@ -85,7 +85,13 @@ def update_profile(fields_dict):
 
         user_updates = {}
         if fields_dict.get("full_name") != user_doc.full_name:
-            user_updates["first_name"] = fields_dict.get("full_name").split(" ")[0]
+            name_parts = fields_dict.get("full_name").split()
+            if len(name_parts) > 1:
+                user_updates["first_name"] = name_parts[0]
+                user_updates["last_name"] = "".join(name_parts[1:])
+            else:
+                user_updates["first_name"] = name_parts[0]
+                user_updates["last_name"] = ""
 
         if fields_dict.get("username") != user_doc.username:
             user_updates["username"] = fields_dict.get("username")
@@ -94,6 +100,8 @@ def update_profile(fields_dict):
             user = frappe.get_doc("User", user_doc.user)
             for field, value in user_updates.items():
                 setattr(user, field, value)
+
+            user.save(ignore_permissions=True)
 
         return True
 
--- a/fossunited/api/profile.py
+++ b/fossunited/api/profile.py
@@ -62,7 +62,6 @@ def update_profile(fields_dict):
         updated_fields = {
             "full_name": fields_dict.get("full_name"),
             "username": fields_dict.get("username"),
-            "route": fields_dict.get("username"),
             "bio": fields_dict.get("bio"),
             "current_city": fields_dict.get("current_city"),
             "about": fields_dict.get("about"),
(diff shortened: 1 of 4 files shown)


fossunited/api/profile.py
```diff
@@ -78,17 +77,23 @@ def update_profile(fields_dict):
...
+            user_updates["first_name"] = fields_dict.get("full_name").split(" ")[0]
 
         if fields_dict.get("username") != user_doc.username:
             user_updates["username"] = fields_dict.get("username")
 
         if user_updates:
-            frappe.db.set_value("User", user_doc.user, user_updates)
+            user = frappe.get_doc("User", user_doc.user)
+            for field, value in user_updates.items():
+                setattr(user, field, value)
```
rahulporuri, line 102: don't we need to call `user.save(...)` here like we do with `profile.save(...)` above? the `setattr` will only modify the `user` in memory but not in the database, right?

fossunited/api/profile.py
```diff
@@ -78,17 +77,23 @@ def update_profile(fields_dict):
...
+        profile = frappe.get_doc(USER_PROFILE, user_doc.name)
+        for field, value in updated_fields.items():
+            if hasattr(profile, field):
+                setattr(profile, field, value)
+        profile.save(ignore_permissions=True)
 
         user_updates = {}
         if fields_dict.get("full_name") != user_doc.full_name:
-            user_updates["full_name"] = fields_dict.get("full_name")
+            user_updates["first_name"] = fields_dict.get("full_name").split(" ")[0]
```
rahulporuri: it's not clear why we only update `first_name` when `full_name` is provided

fossunited/api/profile.py
```diff
@@ -78,17 +77,23 @@ def update_profile(fields_dict):
             "mastodon": fields_dict.get("mastodon"),
         }
 
-        frappe.db.set_value(USER_PROFILE, user_doc.name, updated_fields)
+        profile = frappe.get_doc(USER_PROFILE, user_doc.name)
+        for field, value in updated_fields.items():
+            if hasattr(profile, field):
+                setattr(profile, field, value)
+        profile.save(ignore_permissions=True)
```
rahulporuri, line 84: not sure why this pattern is better than the previous pattern

rahulporuri (commented):
@harshtandiya could we please add unittests here? This is a risky change and it's better to unittest or at least manually test these changes. If you have already manually tested them, could you please add those details to the PR?

Also, you mentioned how the migration can happen - have you tested that migration locally? If you have a the migration script ready, I am happy to test it locally to verify that the route changes as expected.

fossunited/api/profile.py
harshtandiya: Do you mean the `ignore permissions` & `save` method? They enable us to run [document controller methods](https://frappeframework.com/docs/user/en/basics/doctypes/controllers) .

Basically, with previous database api, i would not be able to run any methods that are written [here](https://github.com/fossunited/fossunited/blob/develop/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py) but with document api, we can run those methods as they were originally intended : whether on save, or on update of any info

fossunited/api/profile.py
harshtandiya: fixed this to update first and last name

fossunited/api/profile.py
harshtandiya: Fixed this. It is indeed needed

harshtandiya:
@rahulporuri migration can happen using the `Bulk Update` feature of frappe

fossunited/api/profile.py
rahulporuri: it will be good to add unittests to catch these issues. if you don't have bandwidth at the moment, please open an issue regarding adding unittests for these functions/functionality

rahulporuri (approved):

PR #694, opened 2024-11-12
fix(test): change `throw` -> `raise` for event checkin method

harshtandiya:

- [x] 🐛Bug Fix

## Description
- Attempts to fix the failing test for attendee checkin. The test should raise a `ValidationError` when an already checked-in attendee is tried to be checked in again. 
- However, CI is failing for some days now, because the ValidationError is not being raised.
- Problem ? This tests pass on local setup, perhaps something wrong with CI then?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,6 @@ jobs:
         working-directory: /home/runner/frappe-bench
         run: |
           bench --site test_site set-config allow_tests true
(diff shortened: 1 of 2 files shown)


harshtandiya:
@rahulporuri can you try to run tests on your local? make sure you are on latest of Frappe and this app. Also make sure you have migrated `bench --site <sitename> migrate`

harshtandiya:
Closing as the tests seem to magically pass now without any change 🤷🏽

PR #703, opened 2024-12-04
Make tests better

### LINKED ISSUE #706: Create `fossunited.test.utils`
### Issue description

Many tests have repeating code blocks which can be extracted as methods to be reused. This will reduce code and make for better code readability

harshtandiya:

- [x] 🧑‍💻Refactor

## Description

This PR creates a tests.utils py file, which is supposed to have methods that are most repeated on tests. 

Most repeated test actions:
- Creating a chapter doc
- creating an event doc
- creating an rsvp form doc
- creating rsvp submission doc

This PR also refactors the said repeating code blocks to use the methods from the new utils file.

## Related Issues & Docs
closes #706

### DIFF
--- a/fossunited/chapters/doctype/foss_chapter/test_foss_chapter.py
+++ b/fossunited/chapters/doctype/foss_chapter/test_foss_chapter.py
@@ -1,54 +1,19 @@
-# Copyright (c) 2023, Frappe x FOSSUnited and Contributors
-# See license.txt
-
 import frappe
 from faker import Faker
 from frappe.tests import IntegrationTestCase
 
-from fossunited.doctype_ids import CHAPTER, CITY_COMMUNITY, USER_PROFILE
+from fossunited.doctype_ids import CHAPTER, USER_PROFILE
+from fossunited.tests.utils import generate_test_chapter
 
 fake = Faker()
 
 
 class TestFOSSChapter(IntegrationTestCase):
     def setUp(self):
-        chapter = frappe.get_doc(
-            {
-                "doctype": CHAPTER,
-                "chapter_name": fake.text(max_nb_chars=40),
-                "chapter_type": CITY_COMMUNITY,
-                "slug": fake.slug(),
-                "city": "Pune",
-                "country": "India",
-                "email": fake.email(),
-                "facebook": fake.url(),
-                "instagram": fake.url(),
-                "linkedin": fake.url(),
-                "mastodon": fake.url(),
-                "matrix": fake.url(),
-                "state": "Maharashtra",
-                "x": fake.url(),
-            }
-        )
-        chapter.insert()
-
-        if not frappe.db.exists("Role", "Chapter Team Member"):
-            frappe.get_doc({"doctype": "Role", "role_name": "Chapter Team Member"}).insert(
-                ignore_permissions=True
-            )
-        if not frappe.db.exists("Role", "Chapter Lead"):
-            frappe.get_doc({"doctype": "Role", "role_name": "Chapter Lead"}).insert(
-                ignore_permissions=True
-            )
-
-        lead_profile = frappe.get_doc(USER_PROFILE, {"user": "test@example.com"})
-        chapter.append("chapter_members", {"chapter_member": lead_profile.name, "role": "Lead"})
-        chapter.save()
-
-        self.chapter = chapter
+        self.chapter = generate_test_chapter()
 
     def tearDown(self):
-        frappe.delete_doc(CHAPTER, self.chapter.name)
+        frappe.delete_doc(CHAPTER, self.chapter.name, force=1)
 
     def test_role_assignment_on_create(self):
         # Given a chapter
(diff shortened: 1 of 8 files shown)


fossunited/tests/utils.py
```diff
@@ -0,0 +1,341 @@
...
+        )
+    ```
+    """
+    try:
+        # Validate and get or generate chapter
+        if not chapter:
+            try:
+                chapter = generate_test_chapter(**kwargs)
+            except ImportError:
+                raise ValueError("Unable to generate test chapter")
```
rahulporuri: design comment - it's best not to do such nested calls. it's not too verbose to have 

```
def setUp(self):
    chapter = generate_test_chapter(...)
    event = generate_test_event(chapter=chapter)
    ...
```

in fact, it's best to be explicit and verbose in the test setup.

fossunited/chapters/doctype/foss_chapter/test_foss_chapter.py
```diff
@@ -1,54 +1,19 @@
...
-        lead_profile = frappe.get_doc(USER_PROFILE, {"user": "test@example.com"})
-        chapter.append("chapter_members", {"chapter_member": lead_profile.name, "role": "Lead"})
-        chapter.save()
-
-        self.chapter = chapter
+        self.chapter = generate_test_chapter()
 
     def tearDown(self):
-        frappe.delete_doc(CHAPTER, self.chapter.name)
+        frappe.delete_doc(CHAPTER, self.chapter.name, force=1)
```
rahulporuri: nit - can we use `force=True`? If `1` is a proxy for truthy-ness of passed argument, i would prefer passing a boolean directly. it's a best practice in python that i'm aware of.

fossunited/tests/utils.py
```diff
@@ -0,0 +1,341 @@
...
+    EVENT,
+    EVENT_RSVP,
+    RSVP_RESPONSE,
+    USER_PROFILE,
+)
+
+fake = Faker()
+
+
+def generate_test_chapter(**kwargs):
```
rahulporuri: nitpicky naming comment - Can we rename these to `insert_` instead of `generate_`? `generate_` or `create_` usually mean that something will be returned that the caller has to handle. But these functions insert the relevant object into the database before returning them.

fossunited/chapters/doctype/foss_chapter/test_foss_chapter.py
rahulporuri: i've noticed the same pattern in a bunch of places - both in this PR and elsewhere in the codebase already. we can either choose to address these changes in this PR or make the `force=True` change across the codebase in a single separate PR

fossunited/chapters/doctype/foss_event_rsvp_submission/test_foss_event_rsvp_submission.py
```diff
@@ -52,48 +32,26 @@ def test_unpublish_on_max_count(self):
...
-                "im_a": "Student",
-            }
-        ).insert()
+        frappe.set_user(WEBSITE_USER)
+        generate_rsvp_submission(linked_rsvp=self.rsvp.name, email=WEBSITE_USER)
 
         # Then the email should be added to an email group linked to event for participants
         email_group = frappe.db.get_value(
-            "Email Group", {"event": form.event, "group_type": "Event Participants"}
+            "Email Group", {"event": self.rsvp.event, "group_type": "Event Participants"}
```
rahulporuri, line 50: nit - why not just `self.event` or `self.event.name` (whichever the right thing is)? why `self.rsvp.event`?

fossunited/ticketing/doctype/foss_event_ticket/test_foss_event_ticket.py
```diff
@@ -1,74 +1,24 @@
...
-            }
-        )
-        event.insert(ignore_permissions=True)
-
-        self.event = event
 
     def tearDown(self):
         frappe.set_user("Administrator")
         frappe.delete_doc(EVENT, self.event.name, force=True)
+        frappe.delete_doc(CHAPTER, self.chapter.name, force=True)
```
rahulporuri, line 21: we're deleting `event` and then `chapter` here but earlier in this PR, we were doing `chapter` and then `event`. which is the right order?

`chapter` first and then `event` makes sense to me because `event` depends on `chapter` if i remember correctly.

again, good to use a consistent convention across the codebase - you never know which testcase will be copied to create a new one, worsening this problem in the future.

rahulporuri (commented):
Mostly LGTM but I have a bunch of nitpicky comments that will be good to address.

fossunited/ticketing/doctype/foss_event_ticket/test_foss_event_ticket.py
harshtandiya: doesn't really matter, since its all being force deleted

fossunited/chapters/doctype/foss_event_rsvp_submission/test_foss_event_rsvp_submission.py
harshtandiya: no reason whatsover, just went with this because all of the above give the same value

fossunited/tests/utils.py
harshtandiya: clarifying: we want to remove chapter creation in event creation util method right?

fossunited/tests/utils.py
harshtandiya: in my opinion, generate feels better 😅 the way i name methods is largely influenced by this document: https://gist.github.com/18alantom/d9f0565c0f42d6a71311d4a3093a1331#functions

If insert is the better way to do naming in this instance, by all means we should go ahead and do it

rahulporuri (approved):
Reminder to squash and merge like we discussed

PR #713, opened 2024-12-11
add: test for email workflow

### LINKED ISSUE #712: [BUG] Email portal not working
### Describe the bug

Emailing portal seems to be not working. 
Whenever an organizer is trying to send emails, it throws an error such as this
```
frappe.exceptions.PermissionError: You are not authorised for this action

Uncaught (in promise) Error: /api/method/fossunited.api.emailing.send_test_email PermissionError
```

This is a bug, as the chapter members should have the necessary permissions to send newsletters.

harshtandiya:

- [x] 🐛Bug Fix
- [x] 🧑‍💻Refactor

## Description

The email portal was failing, not sending the emails and throwing a PermissionError. This is an unexpected bug, as the chapter members should have enough permissions to send emails.

The bug seems to occur when the required methods try to run `check_if_chapter_member` method. This method **Should** be returning true for chapter members, but in this instance it was returning false. This is fixed by passing the session user as params.

Additionally: Tests have been added to test the crucial emailing methods, so as to ensure that they work as intended

 
## Related Issues & Docs
closes #712

### DIFF
--- a/fossunited/api/emailing.py
+++ b/fossunited/api/emailing.py
@@ -375,7 +375,7 @@ def send_campaign(campaign_id: str):
     else:
         chapter = campaign.chapter
 
-    if not check_if_chapter_member(chapter=chapter):
+    if not check_if_chapter_member(chapter, frappe.session.user):
         frappe.throw("You are not authorised for this action", frappe.PermissionError)
 
     campaign.flags.ignore_permissions = 1
@@ -400,7 +400,7 @@ def send_test_email(campaign_id: str, email: str):
     else:
         chapter = campaign.chapter
 
-    if not check_if_chapter_member(chapter=chapter):
+    if not check_if_chapter_member(chapter, frappe.session.user):
         frappe.throw("You are not authorised for this action", frappe.PermissionError)
 
     campaign.flags.ignore_permissions = 1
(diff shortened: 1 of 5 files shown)


fossunited/doctype_ids.py
```diff
@@ -44,3 +44,10 @@
 RAZORPAY_PAYMENT = "Razorpay Payment"
 RAZORPAY_SETTINGS = "Razorpay Settings"
 RAZORPAY_WEBHOOK_LOG = "Razorpay Webhook Log"
+
+
+# Emailing
+CAMPAIGN = "Newsletter"
+NEWSLETTER = "Newsletter"
```
rahulporuri: why two different names for the same thing?

fossunited/api/emailing.py
```diff
@@ -375,7 +375,7 @@ def send_campaign(campaign_id: str):
     else:
         chapter = campaign.chapter
 
-    if not check_if_chapter_member(chapter=chapter):
+    if not check_if_chapter_member(chapter, frappe.session.user):
```
rahulporuri, line 376: let's update the function signature to not have a default value for the `user` kwarg

fossunited/doctype_ids.py
rahulporuri: i would prefer using "Campaign" over "Newsletter"

rahulporuri (approved):

PR #714, opened 2024-12-11
feat: automate event conclusions

### LINKED ISSUE #711: [FEATURE] Automate event conclusion
### Feature description

automate event status changing to concluded after the day of event. This can be done using [frappe background jobs](https://docs.frappe.io/framework/user/en/api/background_jobs#scheduler-events)

harshtandiya:

- [x] 🍕Feature

## Description

This PR automates the process of event conclusion at the end of the day / start of new day, when event_end_date < now date.

This uses the [Background Jobs](https://docs.frappe.io/framework/user/en/api/background_jobs) functionality of frappe. Specifically the [Scheduler events](https://docs.frappe.io/framework/user/en/api/background_jobs#scheduler-events).

Added a test for testing the functionality of the method involved. Used the `mock` and `patch` methods by Python's `unittest` library to mock the now_dates to a different dates in future.

## Related Issues & Docs
closes #711

### DIFF
--- a/fossunited/hooks.py
+++ b/fossunited/hooks.py
@@ -100,3 +100,11 @@ override_doctype_class = {
 # Migration
 # ----------------
 after_migrate = "fossunited.migrate.after_migrate"
+
+# Scheduler Events
+# ----------------
+scheduler_events = {
+    "daily_long": [
+        "fossunited.scheduled_tasks.conclude_events",
+    ],
+}
--- a/fossunited/scheduled_tasks.py
+++ b/fossunited/scheduled_tasks.py
@@ -1,0 +1,29 @@
+from datetime import datetime
+
+import frappe
+
+from fossunited.doctype_ids import EVENT
+
+
+def conclude_events():
+    """
+    Get all the events which have ended (end_date < today) and set their status to concluded
+    """
+    events = frappe.db.get_all(
+        EVENT,
+        {"status": "Live", "event_end_date": ["<", datetime.today()]},
+        ["name", "status", "event_end_date", "event_start_date"],
+        page_length=999,
+    )
+
+    for event in events:
+        doc = frappe.get_doc(EVENT, event.name)
+        doc.status = "Concluded"
+        try:
+            doc.save(ignore_permissions=True)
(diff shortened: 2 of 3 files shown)


harshtandiya:
The failing checkin test is a false alarm.

fossunited/hooks.py
```diff
@@ -100,3 +100,11 @@
 # Migration
 # ----------------
 after_migrate = "fossunited.migrate.after_migrate"
+
+# Scheduler Events
+# ----------------
+scheduler_events = {
+    "daily_long": [
```
rahulporuri, line 107: Does this run at the beginning of the day e.g. 00:01 hours or does it run at the end of the day e.g. 23:59 hours? Because this isn't clear from the docs

looking at the `conclude_events` function, it seems to be run at 00:01 hours, is that correct? and the events from yesterday/earlier will be updated today.

rahulporuri (approved):

fossunited/scheduled_tasks.py
```diff
@@ -0,0 +1,29 @@
...
+from fossunited.doctype_ids import EVENT
+
+
+def conclude_events():
+    """
+    Get all the events which have ended (end_date < today) and set their status to concluded
+    """
+    events = frappe.db.get_all(
+        EVENT,
+        {"status": "Live", "event_end_date": ["<", datetime.today()]},
```
rahulporuri, line 14: note : there are events in the platform that don't have an `event_end_date`. I think it isn't a mandatory attribute on the doctype so external events don't seem to have them all the time. Additionally, the old/migrated events don't seem to have this attribute. What happens to those events?

I think that this function will leave those events alone. Right?

harshtandiya: Yes, it will not fetch those

fossunited/hooks.py
harshtandiya: I think it runs at sharp 00:00

Deducing it from frappe's tests: https://github.com/frappe/frappe/blob/develop/frappe/tests/test_scheduler.py

Issue #727, opened 2024-12-17
Add tests for hackathon workflow


harshtandiya:
### Issue description

Add tests for the following doctypes
- [x] Hackathon
- [x] Hackathon Team
- [x] Hackathon Join Team Request
- [x] Hackathon Participant
- [x] Localhosts
- [x] Localhosts Organizer

PR #728, opened 2024-12-18
test: add tests for hackathon & hackathon team doctype

harshtandiya:

- [x] Test

## Description

- Add basic tests for Hackathon & Hackathon Team Doctype
- Add re-useable hackathon related test utility functions

## Related Issues & Docs

partial #727

### DIFF
--- a/fossunited/foss_hackathon/doctype/foss_hackathon/test_foss_hackathon.py
+++ b/fossunited/foss_hackathon/doctype/foss_hackathon/test_foss_hackathon.py
@@ -15,9 +15,9 @@ class TestFOSSHackathon(IntegrationTestCase):
 
     def test_hackathon_route(self):
         if self.hackathon.permalink:
-            self.assertTrue(self.hackathon.route == f"hack/{self.hackathon.permalink}")
+            self.assertEqual(self.hackathon.route, f"hack/{self.hackathon.permalink}")
         else:
-            self.assertTrue(
-                self.hackathon.route
-                == f"hack/{self.hackathon.hackathon_name.lower().replace(' ', '-')}"
+            self.assertEqual(
+                self.hackathon.route,
+                f"hack/{self.hackathon.hackathon_name.lower().replace(' ', '-')}",
             )
--- a/fossunited/foss_hackathon/doctype/foss_hackathon/test_foss_hackathon.py
+++ b/fossunited/foss_hackathon/doctype/foss_hackathon/test_foss_hackathon.py
@@ -1,0 +1,23 @@
+import frappe
+from frappe.tests import IntegrationTestCase
+
+from fossunited.doctype_ids import HACKATHON
+from fossunited.tests.utils import insert_test_hackathon
+
+
+class TestFOSSHackathon(IntegrationTestCase):
+    def setUp(self):
+        self.hackathon = insert_test_hackathon()
+
+    def tearDown(self):
+        frappe.set_user("Administrator")
+        frappe.delete_doc(HACKATHON, self.hackathon.name, force=True)
+
+    def test_hackathon_route(self):
+        if self.hackathon.permalink:
+            self.assertTrue(self.hackathon.route == f"hack/{self.hackathon.permalink}")
+        else:
+            self.assertTrue(
+                self.hackathon.route
+                == f"hack/{self.hackathon.hackathon_name.lower().replace(' ', '-')}"
+            )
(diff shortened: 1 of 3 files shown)


fossunited/foss_hackathon/doctype/foss_hackathon/test_foss_hackathon.py
```diff
@@ -0,0 +1,23 @@
...
+    def setUp(self):
+        self.hackathon = insert_test_hackathon()
+
+    def tearDown(self):
+        frappe.set_user("Administrator")
+        frappe.delete_doc(HACKATHON, self.hackathon.name, force=True)
+
+    def test_hackathon_route(self):
+        if self.hackathon.permalink:
+            self.assertTrue(self.hackathon.route == f"hack/{self.hackathon.permalink}")
```
rahulporuri:
```suggestion
            self.assertEqual(self.hackathon.route, f"hack/{self.hackathon.permalink}")
```

same comment with the other `assertEqual` in this testcase.

fossunited/foss_hackathon/doctype/foss_hackathon_team/test_foss_hackathon_team.py
```diff
@@ -1,60 +1,55 @@
...
-        for _ in range(3):  # since max_team_size is 2, we need 3 participants to test the code
-            participant = frappe.get_doc(
-                {
-                    "doctype": HACKATHON_PARTICIPANT,
-                    "full_name": self.faker.name(),
-                    "email": self.faker.email(),
-                    "hackathon": self.hackathon.name,
-                }
-            ).insert()
+        for _ in range(4):  # since max_team_size is 2, we need 3 participants to test the code
```
rahulporuri: but `max_team_members` is 3 in line 14 and i'm not sure if `max_team_size` is the same as that and you say 3 participants but we iterate 4 times

fossunited/foss_hackathon/doctype/foss_hackathon_team/test_foss_hackathon_team.py
```diff
@@ -1,60 +1,55 @@
...
+            participant.delete(force=True)
+        self.hackathon.delete(force=True)
 
     def test_add_member_to_team(self):
-        for _ in range(self.hackathon.max_team_size):
+        # Given a hackathon with a defined max_team_members size
+        # When the max number of participants are added to that team
+        participant_ids = []
+        for _ in range(self.hackathon.max_team_members):
+            participant_ids.append(self.participants[_].name)
```
rahulporuri: nit - let's not use `_` here and instead just use a standard loop variable like `i` or whatever. `_` is usually meant for things that are disposed of, not to be referred to later.

fossunited/tests/utils.py
```diff
@@ -325,3 +328,123 @@ def insert_rsvp_submission(linked_rsvp: str, **kwargs):
...
+                hackathon_data[key] = value
+
+        hackathon = frappe.get_doc(hackathon_data)
+        hackathon.insert()
+        hackathon.reload()
+
+        return hackathon
+
+    except Exception:
+        raise
```
rahulporuri: not sure why we're doing this - the function will behave exactly this way if we complete remove the `try/except` block. Earlier, we were logging errors so it sort of made sense but i'm not sure it makes sense now - the function will fail loudly so the log messages feel redundant.

what was the intention behind the `try/except` block? this code is meant purely for testing so it's okay to fail fast/clearly.

fossunited/tests/utils.py
```diff
@@ -325,3 +328,123 @@ def insert_rsvp_submission(linked_rsvp: str, **kwargs):
...
+
+
+def insert_test_hackathon_team(hackathon: dict, **kwargs):
+    """
+    Generate a test hackathon team with flexible configuration options.
+
+    Args:
+        hackathon (dict): The hackathon to associate the team with.
+        team_name (str, optional): Name of the team. Defaults to a random name.
```
rahulporuri: `team_name` not explicitly listed in the function definition

rahulporuri (commented):
Mostly LGTM with a few nitpicky-ish comments

fossunited/foss_hackathon/doctype/foss_hackathon_team/test_foss_hackathon_team.py
harshtandiya: Fixed this. I refactored tests but forgot to update this line. We require 4 participants overall to test exceeding member case

fossunited/tests/utils.py
harshtandiya: Just tried to follow the other util functions, but i guess we can lose the try block here. Removed

rahulporuri (approved):

PR #736, opened 2024-12-19
test(hackathon): add tests for `Hackathon Team Join Request` doctype

harshtandiya:

- [x] Test

## Description
Add tests for `Hackathon Team Join Request` doctype
- Test that only a team member can send requests from a specific team.
- Test that acceptance workflow is working
- Made changes to the doctype to define the mandatory fields in a better way

## Related Issues & Docs
partial #727

### DIFF
--- a/fossunited/foss_hackathon/doctype/foss_hackathon_join_team_request/foss_hackathon_join_team_request.py
+++ b/fossunited/foss_hackathon/doctype/foss_hackathon_join_team_request/foss_hackathon_join_team_request.py
@@ -2,11 +2,13 @@
 # For license information, please see license.txt
 
 import frappe
+import frappe.permissions
 from frappe.model.document import Document
 
 from fossunited.doctype_ids import (
     HACKATHON_PARTICIPANT,
     HACKATHON_TEAM,
+    HACKATHON_TEAM_MEMBER,
     JOIN_TEAM_REQUEST,
 )
 
@@ -20,17 +22,19 @@ class FOSSHackathonJoinTeamRequest(Document):
     if TYPE_CHECKING:
         from frappe.types import DF
 
-        hackathon: DF.Link | None
+        hackathon: DF.Link
         hackathon_name: DF.Data | None
         is_outgoing_request: DF.Check
-        reciever_email: DF.Data | None
-        requested_by: DF.Link | None
+        reciever_email: DF.Data
+        requested_by: DF.Link
         sender_name: DF.Data | None
         status: DF.Literal["Pending", "Accepted", "Rejected"]  # noqa: F821
-        team: DF.Link | None
+        team: DF.Link
         team_name: DF.Data | None
     # end: auto-generated types
-    pass
+
+    def before_insert(self):
+        self.validate_sender_as_member()
 
     def before_save(self):
         if self.has_value_changed("status"):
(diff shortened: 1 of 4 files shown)


fossunited/foss_hackathon/doctype/foss_hackathon_join_team_request/foss_hackathon_join_team_request.py
```diff
@@ -60,12 +64,31 @@ def reject_other_requests(self):
...
-            request_doc = frappe.get_doc(JOIN_TEAM_REQUEST, request.name)
+            request_doc = frappe.get_doc(JOIN_TEAM_REQUEST, request)
             request_doc.status = "Rejected"
-            request_doc.save()
+            request_doc.save(ignore_permissions=True)
+
+    def validate_sender_as_member(self):
+        if frappe.permissions.is_system_user(frappe.session.user):
+            return
+
```
rahulporuri, line 88: under what conditions will a system user perform this operation? is this code necessary to handle a testcase or is this really needed for production?

fossunited/foss_hackathon/doctype/foss_hackathon_join_team_request/foss_hackathon_join_team_request.py
```diff
@@ -60,12 +64,31 @@ def reject_other_requests(self):
...
+        participant_id = frappe.db.get_value(
+            HACKATHON_PARTICIPANT, {"user": self.requested_by}, ["name"]
+        )
+
+        is_team_member = frappe.db.exists(
+            HACKATHON_TEAM_MEMBER, {"parent": self.team, "member": participant_id}
+        )
+
+        if not is_team_member:
+            frappe.throw("Request sender is not a team member", frappe.ValidationError)
```
rahulporuri, line 98: i'm not sure how this can happen in a real-world scenario. how does someone send a join request to a user while not being a team member themselves? here again I'm wondering if this code is influenced by the scenario being tested or if it can really happen in production

rahulporuri (commented):
I have a few questions about some of the code that has been added to the doctype.

fossunited/foss_hackathon/doctype/foss_hackathon_join_team_request/foss_hackathon_join_team_request.py
harshtandiya: Backdoor for when some bugs occur and we need to manually do it for someone. Happened a couple of times before, generally its better to give the system user the ability to do it.

Ideally we will never have to use this

fossunited/foss_hackathon/doctype/foss_hackathon_join_team_request/foss_hackathon_join_team_request.py
harshtandiya: It can happen since anyone can create a Join Request, it is just a doctype. We have to make sure that it doesn't through these controller methods

fossunited/foss_hackathon/doctype/foss_hackathon_join_team_request/foss_hackathon_join_team_request.py
rahulporuri: let's document this in the method docstring?

fossunited/foss_hackathon/doctype/foss_hackathon_join_team_request/foss_hackathon_join_team_request.py
rahulporuri: is this scenario possible? From the UI

- ansh, you, and I are registered for foss hack
- you create a team for foss hack
- I send ansh a request to join your team

Having never seen the UI. I'm surprised that there is a way to do this. could we not prevent this in the UI i.e. I can only invite someone to join my team - assuming I created a team in the first place.

harshtandiya: This is not possible through ui, but rather through api calls.

harshtandiya: To prohibit such cases via api calls, this validation is necessary

rahulporuri: I was afraid so. can we please add this info to the docstring here as well?

rahulporuri (approved):
LGTM apart from the comments about adding documentation for the changes in the doctype

harshtandiya:
@rahulporuri docstrings added

PR #743, opened 2024-12-25
test: add tests for Hackathon Localhost doctype

harshtandiya:

- [x] Test

## Description
- Add tests for Hackathon Localhost doctype
- Refactored localhost controller methods, for better readability and performance. Moreover, to make it work as intended. 
- Fixed LocalHost Organizer doctype id, which was wrong in doctype_ids file

## Related Issues & Docs
#727


fossunited/doctype_ids.py
```diff
@@ -7,7 +7,7 @@
 HACKATHON = "FOSS Hackathon"
 JOIN_TEAM_REQUEST = "FOSS Hackathon Join Team Request"
 HACKATHON_LOCALHOST = "FOSS Hackathon LocalHost"
-LOCALHOST_ORGANIZER = "FOSS Hackathon Localhost Organizer"
+LOCALHOST_ORGANIZER = "FOSS Hackathon LocalHost Organizer"
```
rahulporuri, line 10: i guess this wasn't a problem for us because this refactor happened after FOSS Hack 3.0 ended, right?

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/foss_hackathon_localhost.py
```diff
@@ -35,55 +35,70 @@ class FOSSHackathonLocalHost(Document):
...
+        removed_members = [
+            member.profile
+            for member in organizers_old
+            if member.profile not in [m.profile for m in organizers]
+        ]
+
+        self.assign_localhost_organizer_role(new_members)
+        self.remove_organizer_role(removed_members)
+
+    def assign_localhost_organizer_role(self, members: list):
```
rahulporuri:
```suggestion
    def assign_localhost_organizer_role(self, profiles: list):
```

nitpicky comment - the roles are assigned to Profiles so can we rename the arg accordingly? and also update the callers as well because `members` is unclear in our codebase but `profile`/`profiles` has a predefined meaning

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/foss_hackathon_localhost.py
```diff
@@ -35,55 +35,70 @@ class FOSSHackathonLocalHost(Document):
...
+        for member in members:
             user = frappe.get_doc(
                 "User",
-                frappe.db.get_value(USER_PROFILE, member.profile, "user"),
+                frappe.db.get_value(USER_PROFILE, member, "user"),
             )
             user.add_roles("Localhost Organizer")
```
rahulporuri, line 71: this is unrelated to this PR but we should add a constant for this - this is a string that has special meaning so we should make it easily discoverable like the IDs. if there are other roles, we could create a `fossunited.roles` module and add them there. we already define the `Localhost Organizer` string in the `remove_organizer_role` so this feels like a necessary change

nitpicky comment - `LocalHost` in the doctype vs `Localhost` in the role. I wish we had been thorough when we wrote this code originally.

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/foss_hackathon_localhost.py
```diff
@@ -35,55 +35,70 @@ class FOSSHackathonLocalHost(Document):
...
+                frappe.db.get_value(USER_PROFILE, member, "user"),
+            )
+            user.remove_roles("Localhost Organizer")
+            user.save(ignore_permissions=True)
 
-    def other_localhost_member(self, old_member):
+    def is_other_localhost_member(self, member: str) -> bool:
+        """
+        Check if the member is part of any other localhost
+        """
```
rahulporuri:
```suggestion
    def is_other_localhost_member(self, profile: str) -> bool:
        """
        Check if the profile is an organizer of another localhost
        """
```

nitpicky renaming suggestion, similar to an earlier comment

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/foss_hackathon_localhost.py
```diff
@@ -35,55 +35,70 @@ class FOSSHackathonLocalHost(Document):
     pass
 
     def before_insert(self):
-        self.assign_localhost_organizer_role()
+        members = [member.profile for member in self.organizers]
```
rahulporuri: design question : Why do even need the `FOSSHackathonLocalHostOrganizer` doctype? Why not just use `FOSSUserProfile`?

```
organizers: DF.Table[FOSSUerProfile]
```

From the looks of it, the only meaningful info on `FOSSHackathonLocalHostOrganizer` is `FOSSUserProfile` - which surprisingly can be `None` as well which I think might cause subtle bugs.

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/foss_hackathon_localhost.py
```diff
@@ -35,55 +35,70 @@ class FOSSHackathonLocalHost(Document):
...
+        new_members = [
+            member.profile
+            for member in organizers
+            if member.profile not in [m.profile for m in organizers_old]
+        ]
+        removed_members = [
+            member.profile
+            for member in organizers_old
+            if member.profile not in [m.profile for m in organizers]
+        ]
```
rahulporuri: can you think about the time-complexity of this code? we're unnecessarily iterating over loops. how about

```
# There is no point of duplicates here so we could use sets instead of lists
new_profiles = {member.profile for member in organizers}
old_profiles = {member.profile for member in organizers_old}

# using sets makes the comparisons faster
profiles_to_add = list(new_profiles - old_profiles)
profiles_to_remove = list(ld_profiles - new_profiles)
```

please correct me if I fundamentally misunderstood the code

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/foss_hackathon_localhost.py
```diff
@@ -35,55 +35,70 @@ class FOSSHackathonLocalHost(Document):
...
-    def remove_organizer_role(self, old_member):
-        if self.other_localhost_member(old_member):
-            return
-
-        user = frappe.get_doc(
-            "User",
-            frappe.db.get_value(USER_PROFILE, old_member.profile, "user"),
-        )
-        user.remove_roles("Localhost Organizer")
+    def remove_organizer_role(self, members: list):
```
rahulporuri:
```suggestion
    def remove_organizer_role(self, profiles: list):
```

same comment as earlier

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/test_foss_hackathon_localhost.py
```diff
@@ -0,0 +1,169 @@
...
+)
+from fossunited.tests.utils import insert_test_hackathon, insert_test_hackathon_localhost
+
+
+class TestFOSSHackathonLocalHost(IntegrationTestCase):
+    def setUp(self):
+        self.ORGANIZER_1 = "test1@example.com"
+        self.organizer1_profile = frappe.get_doc(USER_PROFILE, {"user": self.ORGANIZER_1})
+        self.ORGANIZER_2 = "test2@example.com"
+        self.organizer2_profile = frappe.get_doc(USER_PROFILE, {"user": self.ORGANIZER_2})
```
rahulporuri, line 16: maybe stupid question - don't we need to save the profiles explicitly? i thought this returns a new document that isn't stored to disk yet.

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/test_foss_hackathon_localhost.py
```diff
@@ -0,0 +1,169 @@
...
+
+        self.localhost.save()
+
+        # Organizer 1 should still have the "Localhost Organizer" Role
+        self.assertTrue(self.has_organizer_role(self.ORGANIZER_1))
+        new_localhost.delete(force=True)
+
+    def has_organizer_role(self, user):
+        roles = frappe.get_roles(user)
+        if "Localhost Organizer" in roles:
```
rahulporuri, line 165: One more reason why it is good to add the `Localhost Organizer` string as a constant somewhere. Not sure what other Roles are added via code similar to this

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/test_foss_hackathon_localhost.py
```diff
@@ -0,0 +1,169 @@
...
+        self.organizer1_profile = frappe.get_doc(USER_PROFILE, {"user": self.ORGANIZER_1})
+        self.ORGANIZER_2 = "test2@example.com"
+        self.organizer2_profile = frappe.get_doc(USER_PROFILE, {"user": self.ORGANIZER_2})
+        self.hackathon = insert_test_hackathon()
+        self.localhost = insert_test_hackathon_localhost(parent_hackathon=self.hackathon.name)
+
+    def tearDown(self):
+        frappe.set_user("Administrator")
+        frappe.delete_doc(HACKATHON_LOCALHOST, self.localhost.name, force=True)
+        self.hackathon.delete(force=True)
```
rahulporuri, line 23: nit - delete the localhost too?
i guess we don't really need to worry about the profiles as long as remove the relevant roles from them, which happens during localhost deletion automatically, right?

rahulporuri (commented):
Mostly LGTM but I have a bunch of nitpicky comments and one potential performance improvement.

I'm loving the tests @harshtandiya because it provides a great opportunity to investigate, and potentially improve/fix old code! Please keep up the work :pray:

fossunited/doctype_ids.py
harshtandiya: Yeah, we did the whole doctype ID thing after FOSS Hack

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/foss_hackathon_localhost.py
harshtandiya: `FOSS Hackathon LocalHost Organizer` is a child doctype, which is just a fancy way to introduce tables to a "parent" doctype. This doctype lets us add a list of profiles to a localhost as organizers
image

I don't think it is possible to just add profile to localhost. You need to create a child table with  profile as a link

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/foss_hackathon_localhost.py
rahulporuri: ignore - i've added a separate issue for this - https://github.com/fossunited/fossunited/issues/744 - to address all of the role constants

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/test_foss_hackathon_localhost.py
harshtandiya: so, test users which i usually write (test1@...., test2@...) are out of box from frappe. 
With our logic for user creation, profiles are auto created when users are created.
We do not need to create profiles for these test users as those are already created when these users are created. 

We just need to `get` them to use in tests. Unless doing explicit update operations

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/test_foss_hackathon_localhost.py
harshtandiya: ah, line 23 is supposed to be like 24, forgot to change

fossunited/foss_hackathon/doctype/foss_hackathon_localhost/foss_hackathon_localhost.py
harshtandiya: This works, changing the code

rahulporuri (approved):
LGTM! Thank you for addressing the review comments

Issue #742, opened 2024-12-25
FOSS Hack Tracker Issue


harshtandiya:
### Issue description

Do not open foss hack 2025 registrations before this issue closes 

Things to do- 

- [x] Hackathon attendees to be added to their own email groups
- [x] Handle Email notifications
- [x] Test workflow

PR #751, opened 2024-12-26
Refactor!: use dynamic `reference_document` in Newsletters, Email Groups doctypes

harshtandiya:

- [x] ⚙️Chore
- [x] 🧑‍💻Refactor

## Description

- Added 2 new fields to `Newsletter`, `Email Group` and `Email Templates`:
    1. `reference_document`
    2. `document_type`

- `Document Type` field is a link field to `DocType` doctype, which will enable us to define what kind of doctype we want to link the 3 email centric doctypes with
- `Reference Document` is a [Dynamic Link](https://docs.frappe.io/erpnext/user/manual/en/dynamic-link-fields) field, which has the option of `document_type`. What this simply means is that this field will link to a document of type defined in `document_type` field.
- This architecture will enable us to make the dynamic linking of Newsletters etc with events/hackathons and more using same field `reference_document`.

This PR also refactors the methods and tests which were using emailing methods to use the latest fields.

This PR also has a patch, which will once after the migration is done when update happens. This patch will simply go through all the Newsletters, Email Groups and Email Templates; and then copy the value of `event` field and set the same value to the `reference_document`. Read more about patches [here](https://docs.frappe.io/framework/user/en/database-migrations) 

Breaking change: this PR makes the `chapter` field in hackathons mandatory. So in future, we will need to make the hackathons to be linked with a chapter.

## Related Issues & Docs
Related #742

### DIFF
--- a/fossunited/api/emailing.py
+++ b/fossunited/api/emailing.py
@@ -135,7 +135,7 @@ def create_newsletter_campaign(
             "doctype": "Newsletter",
             "document_type": document_type,
             "reference_document": reference_document,
-            "chapter": chapter,
+            "chapter": _chapter,
             "sender_name": chapter_dict.chapter_name,
             "sender_email": chapter_dict.email,
             "email_group": recipient_groups,
--- a/fossunited/api/emailing.py
+++ b/fossunited/api/emailing.py
@@ -100,9 +100,10 @@ def create_newsletter_campaign(
     Create a newsletter document linked to the particular event / chapter
 
     Args:
-        data: data to be set in the doctype
-        event: event id
-        chapter: chapter id
+        data(dict): data to be set in the doctype
+        reference_document(str): if of the reference document of type `document_type`
+        document_type(str): type of reference document linked. default: 'FOSS Chapter Event'
+        chapter(str): id of chapter it is linked to
     """
     _reference_document = reference_document
     _chapter = chapter
@@ -161,7 +162,8 @@ def get_newsletter_campaigns(
     Get all newsletter / email campaigns specific to an event or a chapter
 
     Args:
-        event: id of the event
+        reference_document: id of the document linked
+        document_type: doctype of reference_document. default: FOSS Chapter Event
         chapter: id of the chapter
 
     Returns:
@@ -278,7 +280,8 @@ def get_email_groups(
     Get email group for a specific event or chapter
 
     Args:
-        event: id of the event
+        reference_document: id of the document linked
+        document_type: doctype of reference_document. default: FOSS Chapter Event
         chapter: id of the chapter
 
     Returns:
--- a/fossunited/api/emailing.py
+++ b/fossunited/api/emailing.py
@@ -16,31 +16,47 @@ EMAIL_GROUP_TYPES = Literal[
 
 
 def create_email_group(
-    event_id: str,
     type: EMAIL_GROUP_TYPES,
+    reference_document: str,
+    document_type: str = EVENT,
 ):
     """
     Create an email group linked to the event
 
     Args:
-        event: event id
-        session_user: session user id
+        type: type of email group
+        reference_document: id of the reference document of type document_type
+        document_type: type of reference document (default: "FOSS Chapter Event")
     """
-    _event = frappe.get_doc(EVENT, event_id)
+    _doc = frappe.get_doc(document_type, reference_document)
+    _chapter = _doc.get("chapter")
 
     if frappe.db.exists(
-        EMAIL_GROUP, {"event": event_id, "chapter": _event.chapter, "group_type": type}
+        EMAIL_GROUP,
+        {
+            "reference_document": reference_document,
+            "document_type": document_type,
+            "chapter": _chapter,
+            "group_type": type,
+        },
     ):
         raise frappe.ValidationError("Email Group already exists for this event")
 
-    group_title = f"{type}-{event_id}"
+    # This is done to prevent cases when event id and hackathon id are equal
+    group_title = ""
+    if document_type == EVENT:
+        group_title = f"{type}-{reference_document}-Event"
+    else:
+        group_title = f"{type}-{reference_document}-Hackathon"
+
     _group_title = group_title[:140]
     group = frappe.get_doc(
         {
             "doctype": EMAIL_GROUP,
             "title": _group_title,
-            "chapter": _event.chapter,
-            "event": event_id,
+            "chapter": _chapter,
+            "reference_document": reference_document,
+            "document_type": document_type,
             "group_type": type,
         }
     )
@@ -74,7 +90,12 @@ def add_to_email_group(email_group: str, email: str):
 
 
 @frappe.whitelist()
-def create_newsletter_campaign(data: dict, event: str = None, chapter: str = None):
+def create_newsletter_campaign(
+    data: dict,
+    reference_document: str = None,
+    document_type: str = EVENT,
+    chapter: str = None,
+):
     """
     Create a newsletter document linked to the particular event / chapter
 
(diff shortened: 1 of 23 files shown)


harshtandiya:
Rebase added extra commits. I will be squashing these

fossunited/api/emailing.py
```diff
@@ -16,31 +16,47 @@
...
     Args:
-        event: event id
-        session_user: session user id
+        type: type of email group
+        reference_document: id of the reference document of type document_type
+        document_type: type of reference document (default: "FOSS Chapter Event")
     """
-    _event = frappe.get_doc(EVENT, event_id)
+    _doc = frappe.get_doc(document_type, reference_document)
+    _chapter = _doc.get("chapter")
```
rahulporuri, line 32: Won't this make this function only work for `FOSS Chapter Event` and not FOSS Hackathons. `FOSS Hackathon` doesn't have `chapter`, does it?

Edit: Never mind. I just noticed that the hackathon has a `chapter`

fossunited/api/emailing.py
```diff
@@ -16,31 +16,47 @@
...
+            "reference_document": reference_document,
+            "document_type": document_type,
+            "chapter": _chapter,
+            "group_type": type,
+        },
     ):
         raise frappe.ValidationError("Email Group already exists for this event")
 
-    group_title = f"{type}-{event_id}"
+    # This is done to prevent cases when event id and hackathon id are equal
```
rahulporuri, line 45: isn't ID unique e.g. a UUID?

Edit : Never mind. I noticed that event id seems to be a human readable name

fossunited/api/emailing.py
```diff
@@ -74,23 +90,30 @@ def add_to_email_group(email_group: str, email: str):
...
+        chapter(str): id of chapter it is linked to
     """
-    _event = event
+    _reference_document = reference_document
     _chapter = chapter
 
-    if not _event and not _chapter:
-        frappe.throw("Atleast one of event or chapter is required")
+    if not _reference_document and not _chapter:
+        frappe.throw("Atleast one of reference_document or chapter is required")
```
rahulporuri:
```suggestion
        frappe.throw("Either reference_document or chapter need to be provided")
```

fossunited/api/emailing.py
```diff
@@ -74,23 +90,30 @@ def add_to_email_group(email_group: str, email: str):
...
-        event: event id
-        chapter: chapter id
+        data(dict): data to be set in the doctype
+        reference_document(str): if of the reference document of type `document_type`
+        document_type(str): type of reference document linked. default: 'FOSS Chapter Event'
+        chapter(str): id of chapter it is linked to
     """
-    _event = event
+    _reference_document = reference_document
     _chapter = chapter
```
rahulporuri, line 109: unrelated but out of curiosity - why do you use the `_x` convention for some variable names and not others?

fossunited/api/emailing.py
```diff
@@ -110,7 +133,8 @@ def create_newsletter_campaign(data: dict, event: str = None, chapter: str = Non
     newsletter_doc = frappe.get_doc(
         {
             "doctype": "Newsletter",
-            "event": event,
+            "document_type": document_type,
+            "reference_document": reference_document,
             "chapter": chapter,
```
rahulporuri: !!! `chapter` is passed as an argument to the function but `_chapter` is inferred within the function in case `chapter is None` so shouldn't we be using `_chapter` here?

fossunited/foss_hackathon/doctype/foss_hackathon/foss_hackathon.py
```diff
@@ -53,7 +54,6 @@ class FOSSHackathon(WebsiteGenerator):
         is_team_mandatory: DF.Check
         max_team_members: DF.Int
         only_show_logo: DF.Check
-        organizing_chapter: DF.Link | None
```
rahulporuri, line 56: don't we need a migration script for this too? How will existing hackathons work if we make this change in code but don't migrate the database? not sure if I'm missing something

fossunited/patches/v1_0/duplicate_event_name_to_reference_field.py
```diff
@@ -0,0 +1,39 @@
...
+        filters={"event": ("!=", "")},
+        fields=["name", "event"],
+    )
+    for email_template in email_templates:
+        frappe.db.set_value(
+            "Email Template",
+            email_template.name,
+            "reference_document",
+            email_template.event,
+        )
```
rahulporuri, line 21: shouldn't we also be setting `document_Type` for all of them or is it not necessary because of the default that is listed in `fossunited.setup`?

rahulporuri (commented):
One potential bug - apart from that LGTM

fossunited/api/emailing.py
harshtandiya: Good catch, refactoring

fossunited/api/emailing.py
harshtandiya: No particular reason tbh

fossunited/foss_hackathon/doctype/foss_hackathon/foss_hackathon.py
harshtandiya: the existing hackathons don't have chapters linked to them, so this change won't affect anything

fossunited/patches/v1_0/duplicate_event_name_to_reference_field.py
harshtandiya: Default is set to be `FOSS Chapter Event`, and we don't have any hackathon related groups rn so it will be fine. It will auto set it to be FOSS Chapter Event

fossunited/foss_hackathon/doctype/foss_hackathon/foss_hackathon.py
rahulporuri: but because of this, the old hackathons won't render correctly either right? because the `organizing_chapter` attribute disappeared?

harshtandiya: It should not be a problem since i refactored every instance of `organizing_chapter` from templates

fossunited/api/emailing.py
harshtandiya: this was done

fossunited/foss_hackathon/doctype/foss_hackathon/foss_hackathon.py
rahulporuri: I dont think im making myself clear

- there is an old hackathon in the database 
- It has organizing_chapter
- The new changes in this PR use chapter when rendering hackathon page
- wont this crash?

again, maybe im missing something here :/

rahulporuri (approved):
LFTM with one nagging question in my mind which shouldn’t block merging

fossunited/foss_hackathon/doctype/foss_hackathon/foss_hackathon.py
harshtandiya: In templates, it's all conditional statements, so if there's no chapter, nothing will happen.

**The old field will be refactored to `chapter` field.**

Old hackathons have a blank field for `organizing_chapter` they never had any linked chapter to themselves. 

But now the `chapter` field is mandatory, so we will have to create a chapter just for FOSS Hack (like we have done for IndiaFOSS) and link the previous FOSS Hack hackathon and the new one to this.

rahulporuri: > Old hackathons have a blank field for organizing_chapter they never had any linked chapter to themselves.

there we go - that's what I was missing. thank you

harshtandiya:
Merging

PR #753, opened 2024-12-28
Test: add tests for Ticket doctype

harshtandiya:

- [x] Test
- [x] ⚙️Chore

## Description

- added method that checks if the tickets status is `Live` before creating the ticket.
- Added a bunch of necessary tests for ticket doctype

### DIFF
--- a/fossunited/payments/doctype/razorpay_payment/razorpay_payment.py
+++ b/fossunited/payments/doctype/razorpay_payment/razorpay_payment.py
@@ -4,6 +4,7 @@
 import frappe
 from frappe.model.document import Document
 
+from fossunited.doctype_ids import EVENT
 from fossunited.utils.payments import (
     get_in_razorpay_money,
     get_razorpay_client,
@@ -83,3 +84,11 @@ class RazorpayPayment(Document):
     @property
     def is_paid(self):
         return self.status == "Captured"
+
+    def before_insert(self):
+        if self.document_type == EVENT:
+            is_ticket_live = frappe.db.get_value(
+                EVENT, {"name": self.document_name}, "tickets_status"
+            )
+            if not bool(is_ticket_live):
+                frappe.throw("Ticket sale are closed for this event!", frappe.PermissionError)
--- a/fossunited/payments/doctype/razorpay_payment/razorpay_payment.py
+++ b/fossunited/payments/doctype/razorpay_payment/razorpay_payment.py
@@ -87,8 +87,8 @@ class RazorpayPayment(Document):
 
     def before_insert(self):
         if self.document_type == EVENT:
-            is_ticket_live = frappe.db.get_value(
+            ticket_status = frappe.db.get_value(
                 EVENT, {"name": self.document_name}, "tickets_status"
             )
-            if not bool(is_ticket_live):
+            if ticket_status == "Closed":
                 frappe.throw("Ticket sale are closed for this event!", frappe.PermissionError)
--- a/fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
+++ b/fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
@@ -84,11 +84,6 @@ class TestRazorpayPayment(IntegrationTestCase):
         self.event.reload()
 
         # When a payment linked to this event is created
-        payment = insert_test_razorpay_payment(event=self.event.name)
-
-        # Then it should throw an error on attempt to change its status to Captured
+        # Then it should throw an error before it is created
         with self.assertRaises(frappe.PermissionError):
-            payment.status = "Captured"
-            payment.save()
-
-        payment.delete(force=True)
+            insert_test_razorpay_payment(event=self.event.name)
--- a/fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
+++ b/fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
@@ -1,129 +1,94 @@
-import json
-
 import frappe
 from faker import Faker
 from frappe.tests import IntegrationTestCase
 
-from fossunited.doctype_ids import EVENT, RAZORPAY_PAYMENT
+from fossunited.doctype_ids import EVENT_TICKET, RAZORPAY_PAYMENT
+from fossunited.tests.utils import (
+    insert_test_chapter,
+    insert_test_event,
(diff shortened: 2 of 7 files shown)


harshtandiya:
@NagariaHussain , what's the usual workflow of the payment?

We do not want this to happen:
- Payment is done
- Check happens for ticket status happens
- Status is closed, so ticket is not created
- But the payment is taken

rahulporuri (approved):
This looks good to me but another self-review might be good @harshtandiya because this is sensitive code. if @NagariaHussain can also comment, that would be great!

fossunited/payments/doctype/razorpay_payment/razorpay_payment.py
```diff
@@ -4,6 +4,7 @@
 import frappe
 from frappe.model.document import Document
 
+from fossunited.doctype_ids import EVENT
```
NagariaHussain, line 7:
```suggestion
from fossunited.doctype_ids import EVENT_DOCTYPE
```

would read better.

fossunited/payments/doctype/razorpay_payment/razorpay_payment.py
```diff
@@ -83,3 +84,11 @@ def sync_status(self):
     @property
     def is_paid(self):
         return self.status == "Captured"
+
+    def before_insert(self):
+        if self.document_type == EVENT:
+            is_ticket_live = frappe.db.get_value(
+                EVENT, {"name": self.document_name}, "tickets_status"
```
NagariaHussain, line 91:
```suggestion
                EVENT, {"name": self.document_name}, "tickets_status"
```

Why is `tickets` plural here?

NagariaHussain: Also a doesn't read like a name of a boolean/Check field

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
```diff
@@ -1,129 +1,94 @@
...
+        # When a payment linked to this event is created
+        payment = insert_test_razorpay_payment(event=self.event.name)
+
+        # Then it should throw an error on attempt to change its status to Captured
         with self.assertRaises(frappe.PermissionError):
-            self.create_payment_document(self.normal_user)
+            payment.status = "Captured"
+            payment.save()
+
+        payment.delete(force=True)
```
NagariaHussain: This should be take care off in tearDown

fossunited/ticketing/doctype/foss_event_ticket/test_foss_event_ticket.py
```diff
@@ -78,3 +88,59 @@ def test_checkin_as_non_chapter_member(self):
...
+            insert_test_ticket(event=self.event.name)
+
+    def test_ticket_creation(self):
+        # Given an event
+        # When the tickets_status is Live
+        self.event.tickets_status = "Live"
+        self.event.save()
+
+        # Then a ticket for this event should be created without any problem
+        ticket = insert_test_ticket(event=self.event.name)
```
NagariaHussain, line 108: Maybe add an assertion on 1-2 ticket details?

fossunited/ticketing/doctype/foss_event_ticket/test_foss_event_ticket.py
```diff
@@ -78,3 +88,59 @@ def test_checkin_as_non_chapter_member(self):
...
+        # When maximum number of tickets are created of this tier type
+        tickets = []
+        for _ in range(tier.maximum_tickets):
+            _ticket = insert_test_ticket(
+                event=self.event.name,
+                tier=tier.title,
+            )
+            tickets.append(_ticket)
+
+        self.event.reload()
```
NagariaHussain, line 139: You are testing the wrong thing here 😅

NagariaHussain: Insert in bulk?

fossunited/payments/doctype/razorpay_payment/razorpay_payment.py
harshtandiya: This was a conscious choice at the time of creating IDs file

fossunited/ticketing/doctype/foss_event_ticket/test_foss_event_ticket.py
harshtandiya: > You are testing the wrong thing here 😅

Trying to test if the tier disables when hit with max count 😅  Is it incorrectly implemented?

harshtandiya: Not using bulk insert since I want to use the custom util function which handles the field values for the doctype

fossunited/payments/doctype/razorpay_payment/razorpay_payment.py
harshtandiya: Oh this is an error. `tickets_status` is a select. I am supposed to check if it is live in the if condition below

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
harshtandiya: I think testing that on Razorpay Payment capture would make sense, adding multiple ticket creation test there

fossunited/ticketing/doctype/foss_event_ticket/test_foss_event_ticket.py
harshtandiya: I think testing that on Razorpay Payment capture would make sense, adding multiple ticket creation test there

fossunited/payments/doctype/razorpay_payment/razorpay_payment.py
harshtandiya: Made changes for this

fossunited/payments/doctype/razorpay_payment/test_razorpay_payment.py
harshtandiya: Surprisingly, in some cases the teardown fails to remove the instances, and errors happen on other tests. This is one of the reason I prefer to just manually delete when I can.

I shall be making an issue in Framework repo for that.

PR #754, opened 2024-12-29
feat: email workflow for hackathon participants

harshtandiya:

- [x] 🍕Feature
- [x] Test

## Description

- Participant emails are now added to hackathon specific email groups of type "Event Participants"
- Tests added to make code more robust
- `Hackathon` field made mandatory in participant doctype
- Workflow added to create email groups at the time of Hackathon creation.

## Related Issues & Docs
#742 
#727

### DIFF
--- a/fossunited/foss_hackathon/doctype/foss_hackathon_participant/foss_hackathon_participant.py
+++ b/fossunited/foss_hackathon/doctype/foss_hackathon_participant/foss_hackathon_participant.py
@@ -4,6 +4,9 @@
 import frappe
 from frappe.model.document import Document
 
+from fossunited.api.emailing import add_to_email_group, create_email_group
+from fossunited.doctype_ids import EMAIL_GROUP, HACKATHON, HACKATHON_LOCALHOST
+
 
 class FOSSHackathonParticipant(Document):
     # begin: auto-generated types
@@ -17,7 +20,7 @@ class FOSSHackathonParticipant(Document):
         email: DF.Data
         full_name: DF.Data
         git_profile: DF.Data | None
-        hackathon: DF.Link | None
+        hackathon: DF.Link
         is_student: DF.Check
         localhost: DF.Link | None
         localhost_request_status: DF.Literal[
@@ -29,7 +32,8 @@ class FOSSHackathonParticipant(Document):
         wants_to_attend_locally: DF.Check
     # end: auto-generated types
 
-    pass
+    def after_insert(self):
+        self.add_to_email_group()
 
     def before_save(self):
         if self.has_value_changed("wants_to_attend_locally"):
@@ -38,11 +42,49 @@ class FOSSHackathonParticipant(Document):
         if self.has_value_changed("localhost"):
             self.update_request_status()
 
+    def validate(self):
+        if self.wants_to_attend_locally and not self.localhost:
+            frappe.throw("No LocalHost value provided", frappe.ValidationError)
+
+    def add_to_email_group(self):
+        if not frappe.db.exists(
+            EMAIL_GROUP,
+            {
+                "document_type": HACKATHON,
+                "reference_document": self.hackathon,
+                "group_type": "Event Participants",
+            },
+        ):
+            create_email_group(
+                type="Event Participants",
+                reference_document=self.hackathon,
+                document_type=HACKATHON,
+            )
+
+        email_group = frappe.db.get_value(
+            "Email Group",
+            {
+                "reference_document": self.hackathon,
+                "document_type": HACKATHON,
+                "group_type": "Event Participants",
+            },
+            ["name"],
+        )
+
+        add_to_email_group(email_group, self.email)
+
     def update_request_status(self):
         self.localhost_request_status = "Pending"
 
     def handle_localhost_rejection(self):
         if not self.has_value_changed("localhost") and self.localhost_request_status == "Rejected":
+            localhost_name = frappe.db.get_value(
+                HACKATHON_LOCALHOST, self.localhost, "localhost_name"
+            )
+            self.add_comment(
+                "Comment",
+                f"Rejected by localhost: {localhost_name}",
+            )
             self.wants_to_attend_locally = False
 
     def handle_localhost_request(self):
@@ -57,6 +99,8 @@ class FOSSHackathonParticipant(Document):
             return
 
         if (self.localhost == prev_doc.localhost) and self.localhost_request_status == "Rejected":
-            frappe.throw("You have already been rejected from this localhost.")
+            frappe.throw(
+                "You have already been rejected from this localhost.", frappe.PermissionError
+            )
 
         self.localhost_request_status = "Pending"
--- a/fossunited/foss_hackathon/doctype/foss_hackathon_participant/test_foss_hackathon_participant.py
+++ b/fossunited/foss_hackathon/doctype/foss_hackathon_participant/test_foss_hackathon_participant.py
@@ -1,0 +1,112 @@
+import frappe
+from frappe.tests import IntegrationTestCase
+
+from fossunited.doctype_ids import EMAIL_GROUP, EMAIL_MEMBER, HACKATHON, USER_PROFILE
+from fossunited.tests.utils import (
+    insert_test_chapter,
+    insert_test_hackathon,
+    insert_test_hackathon_localhost,
+    insert_test_hackathon_participant,
+)
+
+
+class TestFOSSHackathonParticipant(IntegrationTestCase):
+    def setUp(self):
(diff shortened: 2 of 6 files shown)


fossunited/foss_hackathon/doctype/foss_hackathon_participant/foss_hackathon_participant.py
```diff
@@ -38,11 +42,49 @@ def before_save(self):
...
+
+    def add_to_email_group(self):
+        if not frappe.db.exists(
+            EMAIL_GROUP,
+            {
+                "document_type": HACKATHON,
+                "reference_document": self.hackathon,
+                "group_type": "Event Participants",
+            },
+        ):
```
rahulporuri, line 57: design question : under what condition do we expect this to happen via the UI? a participant shouldn't be able to pick a hackathon that doesn't exist - because all hackathons will now have the email group, right?

just wondering if we can nuke this conditional code.

fossunited/foss_hackathon/doctype/foss_hackathon_participant/foss_hackathon_participant.py
```diff
@@ -38,11 +42,49 @@ def before_save(self):
...
+            create_email_group(
+                type="Event Participants",
+                reference_document=self.hackathon,
+                document_type=HACKATHON,
+            )
+
+        email_group = frappe.db.get_value(
+            "Email Group",
```
rahulporuri, line 65: we're not defining this doctype but it (and the ID) are now central to our workflow. maybe time to add this constant to `doctype_ids`?

fossunited/foss_hackathon/doctype/foss_hackathon_participant/foss_hackathon_participant.py
```diff
@@ -57,6 +99,8 @@ def handle_localhost_request(self):
             return
 
         if (self.localhost == prev_doc.localhost) and self.localhost_request_status == "Rejected":
-            frappe.throw("You have already been rejected from this localhost.")
+            frappe.throw(
+                "You have already been rejected from this localhost.", frappe.PermissionError
```
rahulporuri, line 103: extremely nitpicky comment - rejection sounds a bit harsh. could we go with something like "localhost unable to accommodate you" or something along those lines maybe?

fossunited/foss_hackathon/doctype/foss_hackathon_participant/test_foss_hackathon_participant.py
```diff
@@ -0,0 +1,112 @@
...
+                    ),
+                }
+            ],
+        )
+
+    def tearDown(self):
+        frappe.set_user("Administrator")
+        self.participant_1.delete(force=True)
+        self.hackathon.delete(force=True)
+        self.chapter.delete(force=True)
```
rahulporuri, line 42: and lets delete the localhost too?

rahulporuri (approved):
LGTM with a bunch of minor comments

rahulporuri:
looks like CI isn't happy with one of the tests added in this PR

```
 FAIL  test_localhost_rejection (fossunited.foss_hackathon.doctype.foss_hackathon_participant.test_foss_hackathon_participant.TestFOSSHackathonParticipant)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/frappe-bench/apps/fossunited/fossunited/foss_hackathon/doctype/foss_hackathon_participant/test_foss_hackathon_participant.py", line 111, in test_localhost_rejection
    self = <fossunited.foss_hackathon.doctype.foss_hackathon_participant.test_foss_hackathon_participant.TestFOSSHackathonParticipant testMethod=test_localhost_rejection>
AssertionError: PermissionError not raised
```

fossunited/foss_hackathon/doctype/foss_hackathon_participant/foss_hackathon_participant.py
harshtandiya: Maybe in a different PR

fossunited/foss_hackathon/doctype/foss_hackathon_participant/foss_hackathon_participant.py
harshtandiya: Let's keep this, since we do the same for RSVP and CFP Submissions. 
Also there's the existing 2025 FOSS Hack which can be handled with this. 

We can remove it in a future issue if needed

harshtandiya:
@rahulporuri the tests pass in my local, that is weird

fossunited/foss_hackathon/doctype/foss_hackathon_participant/foss_hackathon_participant.py
harshtandiya: too vague, this case is specific when a person is trying to re-apply to a localhost from which they were "Rejected". Makes error handling better imo 😬

rahulporuri:
> @rahulporuri the tests pass in my local, that is weird

can you try running the test standalone? does it pass when running the full testsuite?

rahulporuri:
> @rahulporuri the tests pass in my local, that is weird

I've just re-run this to see if the failure magically disappears. I'm not sure if there is documentation on debugging frappe apps.

rahulporuri:
@harshtandiya - nope. doesn't look like a random failure. i would recommend trying to debug a little to see if you can figure out what's going on.

harshtandiya:
Can you try running in your local @rahulporuri

harshtandiya:
Wow, found the error. We do not have a workflow (after_install) to create roles. Roles are being created at the time of tests, and they are give the desk access, essentially making them System Users.

Since test1 user is often given roles at multiple tests, it is possible it might have the System access. 

Due to which, in this method in FOSS Hackathon Participant:
```py
    def handle_localhost_request(self):
        prev_doc = self.get_doc_before_save()
        if not prev_doc:
            return

        if frappe.db.get_value("User", frappe.session.user, "user_type") == "System User":
            return

        if not self.wants_to_attend_locally:
            return

        if (self.localhost == prev_doc.localhost) and self.localhost_request_status == "Rejected":
            frappe.throw(
                "You have already been rejected from this localhost.", frappe.PermissionError
            )

        self.localhost_request_status = "Pending"
 ```

The function returns at the second check, so the error was never raised.


I propose a more permanent solution: let's explicitly add a method that creates roles at the time of install and specifies whether the roles have/not have desk access.

rahulporuri:
> I propose a more permanent solution: let's explicitly add a method that creates roles at the time of install and specifies whether the roles have/not have desk access.

This is a testing issue so I'm not sure where you're recommending adding the role creation logic. In the tests?

This is a testing problem - and one that I had inquired about earlier I think. I don't know how well frappe cleans up after itself when running tests so I would prefer not relying heavily on the machinery and always ensure that we're cleaning up after ourselves.

If possible, don't rely on much of the existing machinery and create our own e.g. test accounts to ensure that these problems won't show up in the future.

The fundamental software problem here is the use of what look like singletons from what you describe - objects that are shared across tests - like the database itself. Not cleaning up after ourselves when singletons exist is a recipe for disaster in my humble opinion.

harshtandiya:
The way in which roles are created seem to be a problem too here tbh. We do not have a role creation logic in our codebase at all. So in new machines, the roles are created implicitly at the time of implementation. 

So, imagine in a fresh app, the role `Chapter Team Member` would not exists unless we create a chapter and assign a member to it.
This is troublesome because when Roles are created implicitly, they are given desk access by default.

We have so far not encountered these problems because I always manually created roles in production whenever we introduced new roles on the app.

Creating roles explicitly at the time of install / migrate will let us define that a particular role should not have the desk access

harshtandiya:
The cleanup workflow definitely has its own issues here. One of the reasons I am explicitly writing the tearDown methods and cleaning stuff up manually. Maybe we can raise an issue to the Frappe repo too ?

rahulporuri:
> The cleanup workflow definitely has its own issues here. One of the reasons I am explicitly writing the tearDown methods and cleaning stuff up manually. Maybe we can raise an issue to the Frappe repo too ?

I'm not sure what issue we could create upstream - I think the framework can give us tools to use and not mandate what can be done e.g. give us tools to tag all documents being created in a test and automatically delete them during test teardowm.

Better documentation about the available test objects and their properties/behavior is definitely a meaningful upstream issue - if it doesn't exist already

rahulporuri:
> Creating roles explicitly at the time of install / migrate will let us define that a particular role should not have the desk access

This is clearer to me now - and we should definitely move the role creation logic into the code instead of having it on the desk. Anything that reduces the difference between production and a local dev setup is a win in my book - preventing the need for devs to manually setup things that are essential for the platform.

PR #759, opened 2024-12-31
fix: add custom fields in `before_migrate` hook

harshtandiya:

- [x] ⚙️Chore
- [x] 🐛Bug Fix

## Description

- This will enable the use of updated doctype schemas in patches


harshtandiya:
I should prolly change the method name too

rahulporuri (approved):

PR #773, opened 2025-01-09
fix: rsvp status change error after unpublish

### LINKED ISSUE #772: [BUG] Unable to confirm RSVP applications after RSVP form is unpublished
### Describe the bug

Screenshot from 2025-01-08 23-38-06

### Expected behavior

1. Create an RSVP form
2. Accept RSVP submissions
3. Close the RSVP form while I review submissions
4. Approve/Reject the submissions

### To Reproduce

0. Have access to desk
1. look up RSVPs for 5-year anniversary event
2. attempt to update one of the RSVPs from pending to approved
3. you should see the error mentioned earlier

harshtandiya:

- [x] ⚙️Chore
- [x] 🐛Bug Fix

## Description

- code was doing unnecessary publish check of RSVP on a `validate` method. Moved that check to happen in `before_insert`. Made sure to bypass it for chapter members and system users, in case the chapter members need to add participants from their end.
- Code had a huge bug: we did not create workflow to add email to email groups when host approvals are required. All the RSVPs were auto added to `Event Participant` email group of an event even if their status was `Pending`. Fixed this by actually adding a workflow where a participant is only added to the Email group when their submission status is Accepted.
- Wrote tests for above.

Need to fix: [5 Year anniversary event](https://fossunited.org/c/bengaluru/five-year-anniversary) must have `Event Participant` email group with all the RSVP emails. We need to delete the existing email group members and add accepted participants via desk to that particular email group

## Related Issues & Docs
closes #772

### DIFF
--- a/fossunited/chapters/doctype/foss_event_rsvp_submission/foss_event_rsvp_submission.py
+++ b/fossunited/chapters/doctype/foss_event_rsvp_submission/foss_event_rsvp_submission.py
@@ -1,6 +1,7 @@
 import frappe
 from frappe.model.document import Document
 
+from fossunited.api.chapter import check_if_chapter_member
 from fossunited.api.emailing import add_to_email_group, create_email_group
 from fossunited.doctype_ids import EVENT, EVENT_RSVP, RSVP_RESPONSE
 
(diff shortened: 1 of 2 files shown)


fossunited/chapters/doctype/foss_event_rsvp_submission/foss_event_rsvp_submission.py
```diff
@@ -31,16 +32,38 @@ class FOSSEventRSVPSubmission(Document):
...
+        is_chapter_member = check_if_chapter_member(chapter=self.chapter, user=frappe.session.user)
+
+        # If the user is system user or team member,
+        # don't check for validation before rsvp submission creation
+        if is_system_user or is_chapter_member:
+            return
+
+        rsvp_published = frappe.db.get_value(EVENT_RSVP, self.linked_rsvp, "is_published")
+        if not rsvp_published:
+            frappe.throw("RSVP is not published")
```
rahulporuri, line 65: this is to catch the case where somehow a non-system user and a non-chapter member is trying to submit an RSVP submission to an event where the RSVP form isn't published.

is that correct?

rahulporuri (approved):

fossunited/chapters/doctype/foss_event_rsvp_submission/test_foss_event_rsvp_submission.py
```diff
@@ -119,8 +119,113 @@ def test_invalid_status_at_creation(self):
...
         rsvp.save()
 
-        frappe.set_user(WEBSITE_USER)
+        frappe.set_user("Guest")
         # When a submission is done with status accepted
         # Then a frappe.PermissionError should be raised
         with self.assertRaises(frappe.PermissionError):
             insert_rsvp_submission(linked_rsvp=rsvp.name, status="Accepted")
+
+    def test_status_change_after_unpublish(self):
```
rahulporuri, line 128: could we also test the case where we expect an exception when a user tries to submit to an unpublished RSVP?

fossunited/chapters/doctype/foss_event_rsvp_submission/foss_event_rsvp_submission.py
harshtandiya: Yes

fossunited/chapters/doctype/foss_event_rsvp_submission/test_foss_event_rsvp_submission.py
harshtandiya: @rahulporuri added

PR #778, opened 2025-01-11
fix(event-dashboard): broken event volunteer workflow

### LINKED ISSUE #774: [BUG] Unable to add new member/volunteer for event page
### Describe the bug

```
Uncaught (in promise) TypeError: o.chapter.doc is undefined
```

The error seems to be coming from

https://github.com/fossunited/fossunited/blob/b7260fdb9baaa61a0ba72c27a5375ed891ea7e33/dashboard/src/components/chapter/AddMember.vue#L49

### To Reproduce

Simply try to add a volunteer using the "add new member" button on the event page directly.

harshtandiya:

- [x] 🐛Bug Fix

## Description

- Set default values for `chapter` and `event` props to `AddMember.vue` component to be null. This removes the bug in which the code was trying to make changes to a chapter.member object when chapter was not provided.

## Related Issues & Docs
closes #774

### DIFF
--- a/dashboard/src/components/chapter/AddMember.vue
+++ b/dashboard/src/components/chapter/AddMember.vue
@@ -36,11 +36,11 @@ import { ref, defineProps, defineEmits, computed } from 'vue'
 const props = defineProps({
   chapter: {
     type: Object,
-    default: () => ({}),
+    default: () => null,
   },
   event: {
     type: Object,
-    default: () => ({}),
+    default: () => null,
   },
 })
 


dashboard/src/components/chapter/AddMember.vue
```diff
@@ -36,11 +36,11 @@ import { ref, defineProps, defineEmits, computed } from 'vue'
 const props = defineProps({
   chapter: {
     type: Object,
-    default: () => ({}),
+    default: () => null,
```
rahulporuri, line 39: i wasn't sure if `({})` was null or not.

rahulporuri (approved):
Thank you for the prompt fix

dashboard/src/components/chapter/AddMember.vue
harshtandiya: An empty object is a truthy value apparently. TIL

https://stackoverflow.com/questions/21743921/javascript-is-an-empty-object-a-falsy-one

PR #780, opened 2025-01-11
feat: create custom roles at the time of installation / migration

### LINKED ISSUE #744: [MAINT] Add constants for roles
### Issue description

Two roles seem to be added via the codebase, which aren't easily discoverable

- `CFP Reviewer`
- `Localhost Organizer`
- `Chapter Team Member`
- `Chapter Lead`

Any spelling mistakes in the roles might cause subtle but hard to debug bugs so it might be best to create constants and define them somewhere easily discoverable e.g. `fossunited.roles` similar to `fossunited.ids`.

### LINKED ISSUE #779: Create custom roles at the time of app installation / migration
### Issue description

fossunited app uses certain custom roles for the app. Right now, we had to create these roles via desk for production site. It would be better to automate the creation of custom roles during the app installation & site migrations.

This will make the whole app more robust. 

Moreover, roles are created implicitly at the time of installation / migration right now. The problem with this is, the roles are automatically given desk access by Frappe. We do not want this to happen at all, so creating roles explicitly will help us control the access of different roles 

### Steps to reproduce (if applicable)

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

harshtandiya:

- [x] 🍕Feature
- [x] ⚙️Chore

## Description

Closes #779 

This PR adds the workflow for creating app defined custom roles at the time of installation and migration.

This will ensure that no users with certain roles are given system user access.

I am also assuming that this will fix the random test fails.

Also closes #744 , created a constant file for the Roles. We will now need to move the `doctype_ids.py` file inside the id directory. Maybe also change the name to `doctypes.py`

### DIFF
--- a/fossunited/id/roles.py
+++ b/fossunited/id/roles.py
@@ -1,0 +1,12 @@
+# Storing all the Roles for fossunited app in constants
+
+WEBSITE_USER = "FOSS Website User"
+
+CHAPTER_MEMBER = "Chapter Team Member"
+CHAPTER_LEAD = "Chapter Lead"
+
+REVIEWER = "CFP Reviewer"
+
+HACKATHON_ORGANIZER = "Hackathon Organizer"
+
+LOCALHOST_ORGANIZER = "Localhost Organizer"
--- a/fossunited/install.py
+++ b/fossunited/install.py
@@ -1,13 +1,24 @@
 import click
 
-from fossunited.setup import after_install as setup
+from fossunited.setup import after_install as after_setup
+from fossunited.setup import before_install as before_setup
+
+
+def before_install():
+    try:
+        click.secho("Setting up fossunited...", fg="blue", bold=True)
+        before_setup()
+    except Exception as e:
+        BUG_REPORT_URL = "https://github.com/fossunited/fossunited/issues/new"
+        click.secho("Installation failed for app: fossunited :(", fg="bright_red")
+        click.secho(f"Please try reinstalling the app or report the bug at {BUG_REPORT_URL}")
+        raise e
 
 
 def after_install():
     try:
-        click.secho("Setting up fossunited...", fg="blue")
-        setup()
-        click.secho("App installed successfully!", fg="green")
+        after_setup()
+        click.secho("App installed successfully!", fg="green", bold=True)
     except Exception as e:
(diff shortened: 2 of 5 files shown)


fossunited/id/roles.py
rahulporuri, line 1: Nitpicky comment - why not just have it as fossunited.roles? This is pretty important info and i think its okay keeping it at top level and not everything needs to be in a separate sub package eg fossunited.id

fossunited/install.py
```diff
@@ -1,13 +1,24 @@
...
 
-from fossunited.setup import after_install as setup
+from fossunited.setup import after_install as after_setup
+from fossunited.setup import before_install as before_setup
+
+
+def before_install():
+    try:
+        click.secho("Setting up fossunited...", fg="blue", bold=True)
+        before_setup()
```
rahulporuri, line 10: Q: its not obvious to me why adding roles should be done before setup?

fossunited/migrate.py
```diff
@@ -49,3 +57,14 @@ def has_custom_fields():
             missing_fields[doctype] = doctype_missing_fields
 
     return missing_fields
+
+
+def handle_custom_roles():
+    """
+    Get all the required roles and add them to the app
+    """
+    roles = get_custom_roles()
```
rahulporuri, line 66: I guess we dont need to do a roles check like a fields check because these didnt exist before but in the future if we add a new role, this function will need to be modified, right?

rahulporuri (approved):
Mostly LGTM with a bunch if nitpicky comments and a question

fossunited/id/roles.py
harshtandiya: planning to move `doctype_id` inside this as just `doctype`. Better semantics imo `fossunited.id.roles` & `fossunited.id.doctypes`. You can make out what you are accessing just by the path

fossunited/install.py
harshtandiya: During Install, the framework checks the permissions in doctypes and intrinsically creates any new roles with desk access. We would want to create all the roles beforehand so that it does not happen.

fossunited/migrate.py
harshtandiya: The `create_custom_role` method handles the case if the role is already present

PR #813, opened 2025-02-04
feat(profiles)!: share profile with its linked user

### LINKED ISSUE #767: Perms: share profile with its own user
### Issue description

In Frappe, there's the `User` doctype. This is a very sensitive doctype which should only be accessed by the System User and the "User" itself to which this doctype belongs to. 
Frappe handles the second case by using the **Share** option in Frappe.
image

So a user doctype is shared with the user itself, and given the perms of _Read_ and _Write_.
image

We can leverage this to be used in FOSS Profiles as well. At the time of creation, share the Profile doctype with the user linked to it, and give read and write access.
And for existing profiles, we can just write a patch.

"Sharing" just creates a document of doctype `DocShare`. That looks something like this:

image

Documentation: https://docs.frappe.io/erpnext/user/manual/en/sharing

harshtandiya:

- [x] 🍕Feature
- [x] ⚙️Chore

## Description

Sharing a profile with it's linked user will result in better perms management. This will also help us in removing redundant custom validation code from our codebase.

closes #767 

- This PR adds that functionality. After a profile is created, a method runs which shares the profile with the user linked to itself.
Behind the scenes:
`DocShare` is a doctype that handles the sharing workflow in Frappe. We create `DocShare` document for the profiles, where a user is linked, and the permissions which they should have are defined.

image
_DocShare Schema_

ex: for user `lead@example.com`, suppose there is a profile with id `lead`
Then we would create the following docshare:
```
{
   "user":"lead@example.com",
   "document_type": "FOSS User Profile",
   "document_name": "lead",
   "read": 1,
   "write": 1,
}
```

- This PR also adds a patch, which will handle sharing all the existing profiles in production environment

### DIFF
--- a/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
+++ b/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
@@ -71,6 +71,9 @@ class FOSSUserProfile(WebsiteGenerator):
         self.validate_username()
         self.set_route()
 
+    def after_insert(self):
+        self.share_user_with_self()
+
     def on_update(self):
         prev_user_doc = self.get_doc_before_save()
         if prev_user_doc is None:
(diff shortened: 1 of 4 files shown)


fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
```diff
@@ -133,3 +136,21 @@ def get_context(self, context):
...
+        """
+        share_doc = frappe.get_doc(
+            {
+                "doctype": "DocShare",
+                "user": self.user,
+                "share_doctype": self.doctype,
+                "share_name": self.name,
+                "read": 1,
+                "write": 1,
+                "share": 1,
```
rahulporuri, line 153: q: not sure what the `"share"` permission is for

rahulporuri (approved):

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
harshtandiya: `share` the document with other user

PR #821, opened 2025-02-07
feat: converting and storing profile and banner images in webp in the backend

### LINKED ISSUE #748: [FEATURE] Convert profile pictures to webp after they are uploaded
### Feature description

At the moment, we seem to store the profile pictures uploaded by the users as-is. This is problematic for a number of reasons, one of which is performance of the website, especially for speakers and volunteers who are displayed as part of events.

We might want to convert the pictures from whatever format they are uploaded to `webp` after they are uploaded by the users. The `pillow` Python package enables users to convert a JPEG/PNG /other format image to webp format and store/use the converted webp image instead. See https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#webp

pecpo:

- [x] 🍕Feature
- [x] 🏎️Optimization

## Description

Changed the  set_profile_image and set_cover_image methods to convert and store user uploaded images as webp. This change seems to work with the dashboard page without any faults (have not extensively tested but it seems to work well on my locally hosted version). used the Pillow library as mentioned in #748 for the conversion. 

Converting image into webp with a quality setting of 100, leads to it increasing in size by a lot, so i changed it to 80. Any quality setting above 85 results in the image actually increasing in size. By setting the image quality to 80%, it decreases in size by 20%.

Of note is that on the production website right now it fetches images from github instead of our server if user (logged in?) using github, and these do not seem to be in webp format. Also, there is no compression or checking of image size right now, so this could possibly be abused by users.

## Related Issues & Docs

closes #748

### DIFF
--- a/fossunited/api/profile.py
+++ b/fossunited/api/profile.py
@@ -1,10 +1,11 @@
-import frappe
 import io
+
+import frappe
+from frappe.utils.file_manager import save_file
 from PIL import Image
 
 from fossunited.api.dashboard import get_session_user_profile
 from fossunited.doctype_ids import CHAPTER, RESTRICTED_USERNAME, USER_PROFILE
-from frappe.utils.file_manager import save_file
 
 
 def convert_image_to_webp(image_content: bytes) -> bytes:
@@ -38,11 +39,7 @@ def set_profile_image(file_url: str) -> bool:
         filename = f"profile_{user_doc.name}.webp"
 
         saved_file = save_file(
-            fname=filename,
-            content=webp_image,
-            dt=USER_PROFILE,
-            dn=user_doc.name,
-            is_private=False
+            fname=filename, content=webp_image, dt=USER_PROFILE, dn=user_doc.name, is_private=False
         )
 
         frappe.db.set_value(USER_PROFILE, user_doc.name, "profile_photo", saved_file.file_url)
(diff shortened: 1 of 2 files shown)


rahulporuri:
@pecpo thanks for opening this PR. Please checkout the linter/formatter failures from the CI - these should be straightforward to fix.

Installing and enabling precommit on your machine should prevent these issues from popping up in the future.

fossunited/api/profile.py
```diff
@@ -1,40 +1,93 @@
...
+    """
+    Convert the given image content to WebP format using Pillow.
+    Returns the converted image as bytes.
+    """
+    try:
+        with Image.open(io.BytesIO(image_content)) as img:
+            img = img.convert("RGB")
+            webp_io = io.BytesIO()
+            img.save(webp_io, format="WEBP", quality=80)
+            return webp_io.getvalue()
```
rahulporuri, line 21: did you maybe perform any checks to see if the resultant webp image was smaller than the original image?

pecpo: yes sir it is decreasing, by a margin of roughly 20%. Converting image into webp with a quality setting of 100, leads to it increasing in size by a lot, so i changed it to 80. Any quality setting above 85 results in the image actually increasing in size.

pecpo:
> @pecpo thanks for opening this PR. Please checkout the linter/formatter failures from the CI - these should be straightforward to fix.
> 
> Installing and enabling precommit on your machine should prevent these issues from popping up in the future.

Okay sir will do

harshtandiya (approved):
This works! and the PR is gtg

Things that would be good to do:
- Deleting the original file after converting it to webp
- handle the naming of files in a better way. What happens if the user changes the photo again? do we get a file with the same name as before? Thankfully, it seems that Frappe handles it itself. However, we could have a better way to do this in the code itself. Maybe using the user id and a random hash value? (frappe has inbuild function for this)

- Maybe we can delete the old file when a new file is uploaded as profile photo / cover photo


The PR is gtg by itself. All above are good to have that you can address in this / another PR. Please create an issue for it

fossunited/api/profile.py
rahulporuri: I'll go ahead and resolve this comment here but please update the PR description your response.

pecpo: Done

pecpo:
> * Deleting the original file after converting it to webp

I do not understand, in this code I am only saving the image converted to webp in the db, so what original file do I have to delete?

harshtandiya:
When a person uploads a file, two files are created
1. The original file
2. Webp file of the original

And both of the are saved in the db

Below, both the files are the same. cover file is the webp of the stats png
image

PR #825, opened 2025-02-14
fix: removed duplicate api call

### LINKED ISSUE #809: [QUESTION] Duplicate API calls
### Question or issue

Image
There seems to be 2 duplicate frappe API calls, which are exactly the same. Can we fix this?

pecpo:

- [x] 🐛Bug Fix

## Description

the cause of the duplicate api calls in #809 was that the MyProfile.vue child page and NewAppSidebar.vue component were calling the same API endpoint for userprofile details separately. Lifted the API call up to their common parent (Home.vue) and shared this state info with its children.

## Related Issues & Docs

closes #809

### DIFF
--- a/dashboard/src/components/NewAppSidebar.vue
+++ b/dashboard/src/components/NewAppSidebar.vue
@@ -199,13 +199,7 @@ const props = defineProps({
   },
 })
 
-const user_profile = createResource({
-  url: 'fossunited.api.dashboard.get_session_user_profile',
-})
-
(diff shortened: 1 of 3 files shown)


harshtandiya:
Not the direction we want to go. Also, this will break Sidebar for all the other pages other than the myprofile page.

Also: the fields being fetched in both API calls seem to be different

What we can do: create a global store using [pinia](https://pinia.vuejs.org/) for the user. And feed user data to all other pages on need basis using this store. You can read about it in the docs

harshtandiya:
@pecpo lets close this PR?

PR #853, opened 2025-03-08
fix: handle DuplicateEntryError for ticket emails

harshtandiya:

- [x] 🐛Bug Fix

hotfix. 
We are already handling it for CFP and RSVP, makes sense to handle it for Tickets too.

One might wonder why handle the exception at all if we are going to ignore it in the end, well if we do not handle it explicitly, frappe will throw a ValidationError, which will be even harder to handle.


harshtandiya:
Merging this as a priority

PR #868, opened 2025-03-19
fix: filter enabled tiers in BuyTickets component

harshtandiya:

- [x] 🐛Bug Fix

## Description

There was a HUGE bug in the ticketing page.

So, the event data was being pulled in the ticketing page, and on success, the first tier in the tiers list was selected by default on page load EVEN IF THE TIER WAS DISABLED.

This let people "select" a disabled tier, and buy tickets of a disabled type tier. 

This was fixed by only selecting the `enabled` tiers onSuccess fetching of event, and selecting the first **enabled** tier as a default value.

### DIFF
--- a/dashboard/src/pages/BuyTickets.vue
+++ b/dashboard/src/pages/BuyTickets.vue
@@ -464,7 +464,7 @@ const event = createResource({
     }
   },
   onSuccess(data) {
-    const tiers = data.tiers
+    const tiers = data.tiers.filter((tier) => tier.enabled === 1)
     if (tiers.length > 0) {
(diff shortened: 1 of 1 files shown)


NagariaHussain:
Is it not coming filtered from backend itself? Ideally it should.

harshtandiya:
@NagariaHussain Nope, all the tiers were coming

harshtandiya:
A big redesign PR for tickets is due soon, I will be re-writing a lot of stuff at that moment

harshtandiya:
Merging as priority

PR #870, opened 2025-03-21
made downloadAttendeeList async using webworker

xegtor:
offloaded the csv generation part of download attendee list function to a web worker.


xegtor:
@harshtandiya can you check this pr and tell me if it's alright.

xegtor:
So, In this PR I have created a new file for the web worker that now handles the CSV generation part of the download attendee list function. So now what i have done is the in the main function the `downloadAtendeeList` it now calls the web worker and passes it the data which converts it into a CSV rather than handling the CSV generation on the main thread. In the main function the reason i created the new object for the data because the data object submissions was not clone-able so can't be passed into the web worker. I tested all the functionality of the download button and it behaves the same as before and works the same as well.

PR #899, opened 2025-04-16
[PERFORMANCE] removed unnecessary imports

pecpo:

- [x] 🏎️Optimization

## Description

Unnecessary imports into the website.bundle.js file of functions which are working even without being imported here, empirical testing shows that these functions are already being imported by frappe to be made available to the HTML jinja templates for usage.


pecpo:
@harshtandiya seems like the linter service is broken, can you confirm this?

nk521:
Hey @pecpo, is it possible for you to provide with some metrics?

PR #938, opened 2025-05-12
chore: update `frappe-ui` and `vue` to latest

harshtandiya:


rahulporuri:
not sure what needs to happen with this PR @harshtandiya ?

harshtandiya:
it's error-ing out, needs to be fixed

PR #941, opened 2025-05-14
feat: chapter filter for /events/timeline (closes #940)

### LINKED ISSUE #940: chapter filter for events timeline
will allow user to only view events happening in their city

gi-yt:
- made the search bar case-insensitive
- added a filter that allows users to select their closest chapter
- closes #940 
image
image

### DIFF
--- a/fossunited/fossunited/utils.py
+++ b/fossunited/fossunited/utils.py
@@ -223,34 +223,8 @@ def get_chapter_details():
     Retrieves FOSS Chapter Events and Hackathons, then groups them by month and year, separating
     upcoming and past events.
     """
-    events = frappe.get_all(
-        EVENT,
-        pluck="chapter",
-        filters={
-            "status": ["in", ["Approved", "Live", "Concluded"]],
-            "is_published": 1,
-        },
-        order_by="event_start_date",
-    )
-
-    hackathons = frappe.get_all(
-        HACKATHON,
-        pluck="chapter",
-        filters={
-            "is_published": 1,
-        },
-        order_by="start_date",
-    )
-
-    arr = events + hackathons
-
-    res = []
-
-    for val in arr:
-        if val not in res:
-            res.append(val)
-
-    return res
+    chapters = frappe.db.get_all(CHAPTER, fields=["chapter_name", "name"])
+    return chapters
 
 
 def process_event(event, event_list):
--- a/fossunited/fossunited/utils.py
+++ b/fossunited/fossunited/utils.py
@@ -225,7 +225,7 @@ def get_chapter_details():
     """
     events = frappe.get_all(
         EVENT,
-        pluck="chapter",
+        pluck="chapter_name",
         filters={
             "status": ["in", ["Approved", "Live", "Concluded"]],
             "is_published": 1,
--- a/fossunited/fossunited/utils.py
+++ b/fossunited/fossunited/utils.py
@@ -218,6 +218,41 @@ def get_grouped_events():
     return get_month_grouped_events(events, hackathons)
 
 
+def get_chapter_details():
+    """
+    Retrieves FOSS Chapter Events and Hackathons, then groups them by month and year, separating
+    upcoming and past events.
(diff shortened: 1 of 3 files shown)


fossunited/fossunited/utils.py
```diff
@@ -218,6 +218,41 @@ def get_grouped_events():
...
+        filters={
+            "status": ["in", ["Approved", "Live", "Concluded"]],
+            "is_published": 1,
+        },
+        order_by="event_start_date",
+    )
+
+    hackathons = frappe.get_all(
+        HACKATHON,
+        pluck="chapter",
```
harshtandiya: this will return chapter id not chapter name

harshtandiya:
can you test it with a hackathon event?

fossunited/fossunited/utils.py
gi-yt: nope, it returns in human readable format only.

gi-yt:
> can you test it with a hackathon event?

for hackathon to work properly since it does not have the chapter_name field, I used the actual value of chapter (eg. Mumbai-City Community) rather than the pretty one (eg. MUMBAI)

It is a bit ugly but I'm not sure if there is a fix other than to add chapter_name field to the hackathon doctype, which would be too big of a change for a small feature like this.

What do you suggest I do here?

fossunited/fossunited/utils.py
```diff
@@ -218,6 +218,41 @@ def get_grouped_events():
...
+
+    arr = events + hackathons
+
+    res = []
+
+    for val in arr:
+        if val not in res:
+            res.append(val)
+
+    return res
```
harshtandiya: you can do 
```
chapters = frappe.db.get_all(CHAPTER, {"name": ["in", arr]}, pluck="chapter_name")
```

fossunited/fossunited/utils.py
```diff
@@ -218,6 +218,41 @@ def get_grouped_events():
...
+        order_by="start_date",
+    )
+
+    arr = events + hackathons
+
+    res = []
+
+    for val in arr:
+        if val not in res:
+            res.append(val)
```
harshtandiya: also. why this

fossunited/fossunited/utils.py
```diff
@@ -218,6 +218,41 @@ def get_grouped_events():
...
+    """
+    events = frappe.get_all(
+        EVENT,
+        pluck="chapter",
+        filters={
+            "status": ["in", ["Approved", "Live", "Concluded"]],
+            "is_published": 1,
+        },
+        order_by="event_start_date",
+    )
```
harshtandiya: let's do a db call here frappe.db.get_all
because this might break for guest users

fossunited/fossunited/utils.py
```diff
@@ -218,6 +218,41 @@ def get_grouped_events():
...
+
+    hackathons = frappe.get_all(
```
harshtandiya: db call here too

harshtandiya (changes requested):

fossunited/fossunited/utils.py
gi-yt: this was there to remove duplicates, but if we are getting chapters directly  like you suggested, this isn't necessary

gi-yt:
I have simplified the get_chapter_details code to just blindly give a list of all the chapters.

gi-yt:
@harshtandiya can you review now?

harshtandiya:
Looks good, will review thoroughly in some. Can you just test once by accessing the page as guest user ? i.e. no user logged in

harshtandiya:
Looks good, will review thoroughly in some. Can you just test once by accessing the page as guest user ? i.e. no user logged in

gi-yt:
> Looks good, will review thoroughly in some. Can you just test once by accessing the page as guest user ? i.e. no user logged in

yep, I have checked

harshtandiya (approved):

PR #942, opened 2025-05-14
feat: ability to mark chapter status

### LINKED ISSUE #796: [FEATURE] Mark communities as "defunct", "inactive", "independent"
### Feature description

A number of communities listed on the [city communities page](https://fossunited.org/city-communities) aren't active at the moment (and haven't been active for a significant amount of time). We also have communities (e.g. kochi) which are now independent and it'll be good to mark them as such and redirect/point to their new webpage (e.g. https://kochifoss.org/). "inactive" would be middleground between "active" and "defunct".

Adding the ability to mark communities as such enables other parts of the codebase to treat such communities differently e.g. not list them on the city communities page or list them separately instead.

gi-yt:
closes #796 
image
This should look better in the main instance tho, where all the icons and stuff are there

### DIFF
--- a/dashboard/src/pages/ChapterDetails.vue
+++ b/dashboard/src/pages/ChapterDetails.vue
@@ -80,32 +80,6 @@
           :disabled="true"
           label="Chapter Type"
         />
-        <FormControl
-          v-model="chapter.doc.chapter_status"
-          type="select"
-          :options="[
-            {
-              label: 'Active',
-              value: 'Active',
-            },
-            {
-              label: 'Inactive',
-              value: 'Inactive',
-            },
-            {
-              label: 'Defunct',
-              value: 'Defunct',
-            },
-            {
-              label: 'Independent',
-              value: 'Independent',
-            },
-          ]"
-          variant="subtle"
-          size="md"
-          :disabled="false"
-          label="Chapter Status"
-        />
         <div class="col-span-2">
           <TextEditor
             label="About Chapter"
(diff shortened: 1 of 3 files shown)


dashboard/src/pages/ChapterDetails.vue
```diff
@@ -80,6 +80,32 @@
...
+            {
+              label: 'Independent',
+              value: 'Independent',
+            },
+          ]"
+          variant="subtle"
+          size="md"
+          :disabled="false"
+          label="Chapter Status"
+        />
```
harshtandiya: I dont think we want this to be available to the organizers.

This should be controlled via desk only

harshtandiya (changes requested):

dashboard/src/pages/ChapterDetails.vue
gi-yt: I think organizers should have the ability to change this.

For example, if a chapter is revived by a new member to the team, I dont think its a good idea to have FOSS United Team as a bottleneck.

I dont think there are any security implications of having this available to organizers either

harshtandiya: I think organizers will be removed when a chapter is marked as defunct,

And when a chapter is marked active, the members will be needed to be added again anyway

Unless I am understanding the process differently

gi-yt: This also has the ability of a chapter to be marked as inactive.

I think inactive setting should be controlled by organizers

gi-yt: as discussed on mattermost, I have removed it from dashboard, @harshtandiya is it good to merge?

harshtandiya (approved):
LGTM

PR #944, opened 2025-05-14
feat: embed live stream (closes #771)

### LINKED ISSUE #771: [FEATURE] Embed livestream into an event page
### Feature description

Event pages can now link to a YouTube live stream. It would be great to also embed the live stream into the event page. If the viewers want to interact with the streamers, they could still open the youtube link.

This could be achieved by simply injecting an `iframe` into the event page and dynamically set it to the live stream information provided.

This should be pretty-straightforward. Please see https://github.com/hasgeek/funnel/blob/8b491e34770ff167d8ef2a5ccd916130fc66071f/funnel/templates/macros.html.jinja2#L100-L109 for an example of how HasGeek displays video on the event page!

gi-yt:
closes #771 
image

### DIFF
--- a/dashboard/src/pages/EventDetails.vue
+++ b/dashboard/src/pages/EventDetails.vue
@@ -165,6 +165,13 @@
           label="Livestream Link"
           size="md"
         />
+        <FormControl
+          v-model="event.doc.livestream_embed_link"
+          type="url"
+          label="Livestream Link (For embedding)"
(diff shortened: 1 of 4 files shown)


gi-yt:
@harshtandiya is this part also in the purview of your event redesign?

harshtandiya:
I don't think we have a Livestream section in the redesign. But I guess we can add this for now

Please connect with jeswin and look at the updated designs for chapter and event pages

harshtandiya (approved):
LGTM

PR #945, opened 2025-05-15
fix: dont use grouped view if a city chapter is filtered

gi-yt:
small bug on my part, didnt factor in how many gaps would come in the main deployment
Before:
image
Now:
image


gi-yt:
ah goddammit it included the previous commits as well, ill move it to a new branch and submit a fresh PR

PR #948, opened 2025-05-15
add opengraph for /c/CHAPTER/EVENT route

### LINKED ISSUE #831: [FOSS Hack] Dynamic OG Images in platform
#500 

We want to make a workflow which will enable the generation of dynamic OG images for things such as: City Communities, FOSS Clubs, Events, CFP Pages etc.

The general idea is to use [`satori`](https://github.com/vercel/satori) and maybe [`satori-html`](https://github.com/natemoo-re/satori-html) to achieve this. The setup would require creation of multiple API endpoints that can handle OG image creation in different formats as per the requirement.

The designs are here: https://www.figma.com/design/GBYCtoYLihWSpnhS9EScQu/Open-Graph-Images?node-id=0-1&t=wxHfntb2l53d3JfZ-1

The OG images for things like chapter, events etc are dynamic. What that means is that the design and format are the same but the value for data in the image changes such as event name. A similar example could be how Github does OG images for repos.

After the endpoints are created, we have to figure out a way to use them inside the <head> tag to web templates and pages.

gi-yt:
will eventually close #831 

this is for all dynamic page routes. Makes use of https://github.com/fossunited/ograph-gen.

### DIFF
--- a/fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
+++ b/fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
@@ -1,6 +1,8 @@
 # Copyright (c) 2023, Frappe x FOSSUnited and contributors
 # For license information, please see license.txt
 
+import re
+import textwrap
 from datetime import datetime
 
 import frappe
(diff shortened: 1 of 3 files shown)


harshtandiya:
I feel like we should do this:

- Add a field for og image url in the doctype, and take the value for it from that field. 
- Populate those field via custom server scripts for image fields rather than from code. The reasoning is the same as newsletter

Let me know your thoughts

gi-yt:
> I feel like we should do this:
> 
> * Add a field for og image url in the doctype, and take the value for it from that field.
> * Populate those field via custom server scripts for image fields rather than from code. The reasoning is the same as newsletter
> 
> Let me know your thoughts

Agreed, this was just a temporary thing to make sure it works properly. In the long run I agree, a template would be great and also cut down on duplicate code (like reusing the same image for twitter, or description for image:alt)

gi-yt:
@harshtandiya is this workaround fine by you?

gi-yt:
@harshtandiya can you review?

rahulporuri (approved):

PR #949, opened 2025-05-16
fix: [login] change email input type from text to email

kumawatdarshan:
Searched through a simple regex query, did not find any other instance. I guess this one was an oversight.


harshtandiya:
Interesting, because we don't use this page at all!

This comes as a starter template from frappe-ui.

We need to actually delete Login.vue altogether!

kumawatdarshan:
> Interesting, because we don't use this page at all!
> 
> This comes as a starter template from frappe-ui.


Then what's the component in [this page](https://fossunited.org/login#login)

Checking with devtools it has the `type=text`

harshtandiya:
Frappe's inbuilt login & signup pages @darshanCommits

kumawatdarshan:
> Frappe's inbuilt login & signup pages @darshanCommits 

So gotta take the Changes upstream, alright.

PR #950, opened 2025-05-18
fix: make list of campaigns visible by def

### LINKED ISSUE #936: List of campaigns is only visible after a new campaign is created in same session
In the mailing section of an event, you cannot see the list of pre-existing campaigns until you create a new campaign.

Even then, if you refresh the page, the list of pre-existing campaigns stops being visible again.

https://github.com/user-attachments/assets/7bf33925-2789-4e28-8525-2ef54c65c983

gi-yt:
closes #936

### DIFF
--- a/dashboard/src/pages/mailing/EventMailing.vue
+++ b/dashboard/src/pages/mailing/EventMailing.vue
@@ -69,10 +69,16 @@ const event = createDocumentResource({
   name: route.params.id,
   fields: ['*'],
   auto: true,
-  onSuccess() {
-    campaigns.fetch()
-  },
 })
+
+watch(
+  () => event.doc,
+  (doc) => {
+    if (doc) {
+      campaigns.fetch()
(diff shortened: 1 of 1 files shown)


dashboard/src/pages/mailing/EventMailing.vue
```diff
@@ -69,11 +69,17 @@ const event = createDocumentResource({
...
 
+watch(
+  () => event.doc,
+  (doc) => {
+    if (doc) {
+      campaigns.fetch()
+    }
+  },
+)
+
```
harshtandiya: Let's make this explicitly deep and immediate:

```js
watch(
  () => event.doc,
  (doc) => {
    if (doc) {
      campaigns.fetch()
    }
  },
 { 
    deep: true,
    immediate: true
 }
)
```

harshtandiya (approved):

PR #952, opened 2025-05-19
fix: validate event and tier parent

harshtandiya:
validate that the event ticket tier belongs to the specific event.

Earlier this was not validated, which could enable people to buy tickets for an event with a higher ticket price, on a lower price by passing a false tier of another event with low price.


harshtandiya:
~~false alarm on the failing test~~
Fixed

PR #962, opened 2025-05-21
Sort CFPs by approvability (closes #630)

### LINKED ISSUE #630: [FEATURE] Sort CFPs by approvability on the event dashboard
### Feature description

The CFPs should be sorted by approvability (as per CFP reviews) so it's easier for organisers to directly accept/reject the talks at the extreme ends.

gi-yt:
closes #630


harshtandiya (approved):
This might need refactoring later. We should convert the `positive`, `negative`, `not sure` and `approvability` fields to be virtual in the doctype

PR #963, opened 2025-05-21
feat: include unsureity factor while calculating approvability %

gi-yt:
this is up for discussion.

Through this PR, I am giving unsureity a weightage of 0.5, while negative has -1 and positive has +1


harshtandiya:
Up for discussion. 

@rahulporuri thoughts?

ansharora28:
weights for unsureity will be problematic, since "not sure" in many cases is also just "i'm not an expert in this topic and am skipping the review". 

We're using the standard weight values as done by other review platforms. Can confirm pretalx follows the same system

idlip:
Ok, closing since its quiet here.

Again, someday if someone raised query on whats consideration for this.
We might discuss again to sort out words (some are confusing)

PR #965, opened 2025-05-21
RSVP closed notification to organiser

### LINKED ISSUE #568: Notify the team members of the chapter once rsvp reaches max count
### Feature description

Currently RSVP closes once we reach max count but unless I check I am not aware of it so it would be great If I am notified when we reach a threshold / max count so I can increase the limit as well if needed and I am aware the RSVP form was unpublished

### Use case

_No response_

### Benefits

_No response_

### Alternatives considered

_No response_

### Additional context

_No response_

ansharora28:
Added an email notification to be sent to organisers when RSVP reaches max count

image

closes #568

### DIFF
--- a/fossunited/chapters/doctype/foss_event_rsvp_submission/foss_event_rsvp_submission.py
+++ b/fossunited/chapters/doctype/foss_event_rsvp_submission/foss_event_rsvp_submission.py
@@ -3,7 +3,7 @@ from frappe.model.document import Document
 
 from fossunited.api.chapter import check_if_chapter_member
 from fossunited.api.emailing import add_to_email_group, create_email_group
-from fossunited.doctype_ids import EVENT, EVENT_RSVP, RSVP_RESPONSE
+from fossunited.doctype_ids import CHAPTER, EVENT, EVENT_RSVP, RSVP_RESPONSE
 
 
 class FOSSEventRSVPSubmission(Document):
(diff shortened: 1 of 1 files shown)


fossunited/chapters/doctype/foss_event_rsvp_submission/foss_event_rsvp_submission.py
```diff
@@ -78,6 +78,31 @@ def close_rsvp_on_max_count(self):
                 "is_published",
                 False,
             )
+            self.notify_organizers()
+
+    def notify_organizers(self):
+        # Notify the organizers that the RSVP has reached its maximum count
+        # This can be done via email or any other notification system
+        organizer_email = frappe.db.get_value(
+            CHAPTER,
```
gi-yt, line 87: I think you should use EVENT_VOLUNTEER instead of CHAPTER. In most cases this won't make much of a difference but its still good to take care of an edge-case like this

fossunited/chapters/doctype/foss_event_rsvp_submission/foss_event_rsvp_submission.py
```diff
@@ -78,6 +78,31 @@ def close_rsvp_on_max_count(self):
...
+            filters={"name": self.chapter},
```
gi-yt, line 88: here again, use the event ID instead of chapter ID

gi-yt (dismissed):

fossunited/chapters/doctype/foss_event_rsvp_submission/foss_event_rsvp_submission.py
ansharora28: I think chapter email (city@fossunited.org) is cleaner since that will only need us to send one email (city@fossunited.org) instead of fetching and sending it out to individual volunteers

gi-yt: hm, thats a fair point. but I'm worried about cases where the primary volunteer of an event isnt part of city@fossunited.org email group since they are only managing that specific event/are a new volunteer. @rahulporuri @harshtandiya what do you guys think about this?

harshtandiya:
I think the email should go centrally to the organizing chapter's email rather than volunteers individually

fossunited/chapters/doctype/foss_event_rsvp_submission/foss_event_rsvp_submission.py
ansharora28: The communication flow of the events is entirely based around this email. Any queries,speaker confirmations etc. for this event will go to chapter@fossunited.org so the team should ideally loop the lead organiser or add them to the group anyway.

harshtandiya (approved):

PR #966, opened 2025-05-21
mandatory remarks if CFP is rejected/notsure

### LINKED ISSUE #915: [FEATURE] <Make comment mandatory if CFP status is rejected or not sure>
### Feature description

Currently, for CFP review, if the status of the CFP is changed to Rejected or Not sure, making the reviewer comment mandatory would be useful for CFP raisers. It would help them to understand why their proposal is not approved and will help them to make better CFP's.

### Use case

_No response_

### Benefits

_No response_

### Alternatives considered

_No response_

### Additional context

_No response_

gi-yt:
No screenshots but I have tested it and works perfectly
closes #915

### DIFF
--- a/dashboard/src/components/reviewers/ReviewCommentBox.vue
+++ b/dashboard/src/components/reviewers/ReviewCommentBox.vue
@@ -71,53 +71,61 @@ const reviewerProfile = createResource({
 })
 
 const submitReview = () => {
-  createResource({
-    url: 'frappe.client.insert',
-    makeParams() {
-      return {
-        doc: {
-          doctype: 'FOSS Event CFP Review',
-          parenttype: 'FOSS Event CFP Submission',
-          parent: props.submissionId,
-          parentfield: 'reviews',
-          remarks: remarks.value,
-          to_approve: review.value,
-          reviewer_profile: reviewerProfile.data.name,
-          reviewer: reviewerProfile.data.full_name,
-        },
-      }
-    },
-    auto: true,
-    onSuccess() {
-      emits('add:review')
-    },
-    onError(err) {
-      toast.error('Failed to submit review', err.message)
-    },
-  })
+  if ((review.value === "No" || review.value === "Maybe") && (!remarks.value || remarks.value === "<p></p>")) {
+    toast.error('cannot submit review unless remarks are added')
+  } else {
+    createResource({
+      url: 'frappe.client.insert',
+      makeParams() {
+        return {
+          doc: {
+            doctype: 'FOSS Event CFP Review',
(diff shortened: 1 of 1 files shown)


harshtandiya (commented):
this might not be the ideal way to do this 😅 

Can you check how I do error message handling in other places?

I use ErrorMessage component, 

Also, using a central function would be better here, rather than having 2 implementations of the same thing

harshtandiya:
```vue
<script setup>

const errorMessages = ref('')

const validateRemark = () => {
   const errors = []

   if (remark.value != "Yes" && (!remarks.value || remarks.value == "<p><p>") {
       errors.push("<error message here>")
   }

  <any more validations>

  return errors
}


const submitReview = () => {
    const errors  = validateRemarks()
    if (errors.length) {
          errorMessages.value = errors.join('/n')
          return
    }
   ....
}


... same for the other function

```


good reference to this in ManageSponsorDialog.vue

or anywhere I have used the ErrorMessage component

gi-yt:
I have used the method you suggested from ManageSponsorDialog. @harshtandiya is it good to merge now?

dashboard/src/components/reviewers/ReviewCommentBox.vue
```diff
@@ -17,9 +17,10 @@
       :custom-actions="getCustomAction()"
     />
   </div>
+  <ErrorMessage :message="errorMessages" class="text-sm -mb-4" />
```
harshtandiya: nit: can we more this error message above comment box

harshtandiya:
Can you test? Also add a screenshot if possible

gi-yt:
image
image

dashboard/src/components/reviewers/ReviewCommentBox.vue
```diff
@@ -70,7 +71,23 @@ const reviewerProfile = createResource({
   auto: true,
 })
 
+const errorMessages = ref('')
+
+const validateRemark = () => {
+  const errors = []
+
+  if (review.value != "Yes" && (!remarks.value || remarks.value === "<p></p>")) {
+       errors.push("cannot submit review unless remarks are added")
```
harshtandiya: Can you change the message here:

`You can't submit the review without adding remarks.`

rahulporuri: how about "Remarks need to be provided when rejecting a proposal" or "Please provide remarks when rejecting a proposal"

gi-yt:
image
@harshtandiya is it good to merge now?

harshtandiya (approved):

PR #972, opened 2025-05-23
feat: show conferences that a user has attended in foss profile page

gi-yt:
image
image
image

### DIFF
--- a/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.js
+++ b/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.js
@@ -1,8 +1,0 @@
-// Copyright (c) 2025, Frappe x FOSSUnited and contributors
-// For license information, please see license.txt
-
-// frappe.ui.form.on("FOSS User Profile", {
-// 	refresh(frm) {
-
-// 	},
-// });
--- a/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.js
+++ b/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.js
@@ -1,0 +1,8 @@
+// Copyright (c) 2025, Frappe x FOSSUnited and contributors
+// For license information, please see license.txt
+
+// frappe.ui.form.on("FOSS User Profile", {
+// 	refresh(frm) {
+
+// 	},
+// });
--- a/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
+++ b/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
@@ -12,7 +12,7 @@
   "meta_info_section",
   "is_published",
   "is_private",
-  "show_confs",
+  "show_activity",
   "column_break_jkxk",
   "route",
   "header_section",
@@ -276,9 +276,9 @@
   },
   {
    "default": "0",
-   "fieldname": "show_confs",
+   "fieldname": "show_activity",
    "fieldtype": "Check",
-   "label": "Show conferences you have attended publicly"
+   "label": "Show Community Activity"
   }
  ],
  "has_web_view": 1,
@@ -286,7 +286,7 @@
  "index_web_pages_for_search": 1,
  "is_published_field": "is_published",
  "links": [],
- "modified": "2025-05-23 17:55:17.019179",
+ "modified": "2025-05-26 15:44:50.259551",
  "modified_by": "arya@projectsegfau.lt",
  "module": "FOSS Profiles",
  "name": "FOSS User Profile",
--- a/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
+++ b/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
@@ -1,319 +1,327 @@
 {
-  "actions": [],
-  "allow_guest_to_view": 1,
-  "allow_rename": 1,
-  "autoname": "hash",
-  "creation": "2023-07-03 02:44:46.631269",
-  "default_view": "List",
-  "doctype": "DocType",
-  "editable_grid": 1,
-  "engine": "InnoDB",
-  "field_order": [
-    "meta_info_section",
-    "is_published",
-    "is_private",
-    "column_break_jkxk",
-    "route",
-    "header_section",
-    "profile_photo",
-    "column_break_tnli",
-    "cover_image",
-    "basic_information_section",
-    "user",
-    "username",
-    "gender",
-    "bio",
-    "column_break_yqk3",
-    "full_name",
-    "email",
-    "current_city",
-    "website",
-    "about",
-    "socials_section",
-    "github",
-    "gitlab",
-    "linkedin",
-    "mastodon",
-    "medium",
-    "column_break_hmr6",
-    "x",
-    "instagram",
-    "devto",
-    "youtube",
-    "skills_section",
-    "skills",
-    "employment_status_section",
-    "education",
-    "work_experience_section",
-    "experience",
-    "projects_section",
-    "projects"
-  ],
-  "fields": [
-    {
-      "fieldname": "basic_information_section",
-      "fieldtype": "Section Break",
-      "label": "Basic Information"
-    },
-    {
-      "fieldname": "user",
-      "fieldtype": "Link",
-      "in_list_view": 1,
-      "label": "User",
-      "options": "User",
-      "reqd": 1,
-      "unique": 1
-    },
-    {
-      "fieldname": "username",
-      "fieldtype": "Data",
-      "label": "Username",
-      "reqd": 1,
-      "unique": 1
-    },
-    {
-      "description": "Short bio about yourself",
-      "fieldname": "bio",
-      "fieldtype": "Small Text",
-      "label": "Bio"
-    },
-    {
-      "fieldname": "column_break_yqk3",
-      "fieldtype": "Column Break"
-    },
-    {
-      "fieldname": "full_name",
-      "fieldtype": "Data",
-      "label": "Full Name"
-    },
-    {
-      "fetch_from": "user.user_image",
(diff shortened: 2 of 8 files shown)


rahulporuri:
@gi-yt I would rename the profile toggle to something like "Show Community Activity" or something along those lines in order to consolidate all of the planned improvements under the same flag.

One-flag-per-feature is overkill so we need a generic-sounding flag that captures all of the planned improvements

gi-yt:
> @gi-yt I would rename the profile toggle to something like "Show Community Activity" or something along those lines in order to consolidate all of the planned improvements under the same flag.
> 
> One-flag-per-feature is overkill so we need a generic-sounding flag that captures all of the planned improvements

I think it is better to separate the event, talk proposals and hackathons.

might make it a bit more cluttered but I for one would appreciate the discretion in allowing some but not all of them

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
```diff
@@ -132,6 +135,40 @@ def get_context(self, context):
...
+                        "event_start_date",
+                        "event_name",
+                        "banner_image",
+                        "must_attend",
+                        "event_location",
+                    ],
+                    filters={"name": val.event},
+                    page_length=9999,
+                ):
+                    context.events.append(val2)
```
harshtandiya: 1. Let's extract this logic into a separate method

harshtandiya: check ins don't happen for non-paid events

harshtandiya: In general, let's not check for check-ins. Rather if they have RSVPd for the event or bought a ticket for an event.

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
```diff
@@ -132,6 +135,40 @@ def get_context(self, context):
                 experiences_dict[experience.company] = []
             experiences_dict[experience.company].append(experience.as_dict())
         context.experiences_dict = experiences_dict
+
+        events = frappe.db.get_all(
+            EVENT_TICKET,
+            fields=["event", "name"],
+            filters={"email": self.email},
+            page_length=9999,
+        )
```
harshtandiya: you shoudldn't be getting the "name" field here

use `pluck` instead of fields if you are pulling only 1 field, in this case `pluck = "event"`

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
```diff
@@ -132,6 +135,40 @@ def get_context(self, context):
...
+
+        rsvp = frappe.db.get_all(
+            RSVP_RESPONSE,
+            fields=["event", "name", "im_a"],
+            filters={"email": self.email},
+            page_length=9999,
+        )
```
harshtandiya: I think we only need `event` here too

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
harshtandiya, line 1: can you manually save this file and see if the formatting goes back to normal? for some reason this is not being formatted correctly

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
```diff
@@ -1,319 +1,327 @@
...
+   "fieldname": "is_private",
+   "fieldtype": "Check",
+   "label": "Is Private?"
+  },
+  {
+   "default": "0",
+   "fieldname": "show_confs",
+   "fieldtype": "Check",
+   "label": "Show conferences you have attended publicly"
+  }
```
harshtandiya: `show_activity` rather than `show_confs`

Also: don't truncate variable names man, we should be as explicit as possible

fossunited/foss_profiles/doctype/foss_user_profile/templates/foss_user_profile.html
```diff
@@ -188,6 +189,24 @@ <h5>{{ education.institution }}</h5>
...
+      <h3 class="subtitle">Events Attended</h3>
+      {% if doc.show_confs %}
+      <div class="events-grid-4">
+        {% for event in events %}{{ event_card(event) }}{% endfor %}
+      </div>
+      {% else %}
+      <div>User has chosen to keep their attended events anonymous</div>
+      {% endif %}
+      </div>
+  </div>
```
harshtandiya: I hate how this looks 😅 but we don't have anything design-wise. 

Let's go with this rn and ask jeswin to redesign

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
gi-yt: > check ins don't happen for non-paid events

yes, that is why i added `or val.im_a` since im_a variable is only available on non-paid (ie. RSVP) events

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
gi-yt: need the name to check if the person checked in

fossunited/foss_profiles/doctype/foss_user_profile/templates/foss_user_profile.html
gi-yt: agreed

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
harshtandiya: not an ideal flow

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
harshtandiya: Let's not do check ins

gi-yt: alr

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
gi-yt: done

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.json
gi-yt: done

gi-yt:
@harshtandiya can you do a final review, I have applied all changes you have suggested and also re-tested.

@rahulporuri have also renamed it to "Show Community Activity" and a corresponding description on the dashboard
image
image

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.js
rahulporuri, line 1: let's delete this file - it does nothing

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
```diff
@@ -18,25 +19,25 @@ class PrivateProfileError(PermissionError):
 
 class FOSSUserProfile(WebsiteGenerator):
     # begin: auto-generated types
+    # ruff: noqa
```
harshtandiya: disables for the whole file, please revert

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
```diff
@@ -119,6 +122,42 @@ def validate_username(self):
     def set_route(self):
         self.route = f"u/{self.username}"
 
+    def get_user_activity(self):
+        activity = []
+
+        events = frappe.db.get_all(
```
harshtandiya: paid_event_ids =

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
```diff
@@ -119,6 +122,42 @@ def validate_username(self):
...
+            EVENT_TICKET,
+            pluck="event",
+            filters={"email": self.email},
+            page_length=9999,
+        )
+
+        rsvp = frappe.db.get_all(
```
harshtandiya: rsvpd_event_ids =

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
```diff
@@ -119,6 +122,42 @@ def validate_username(self):
...
+                    "event_start_date",
+                    "event_name",
+                    "banner_image",
+                    "must_attend",
+                    "event_location",
+                ],
+                filters={"name": val},
+                page_length=9999,
+            ):
+                activity.append(val2)
```
harshtandiya: ?

why double loops?

event_ids = paid_event_ids + rsvpd_event_ids
for event_id in event_ids:
    event = frappe.db.get_value(EVENT, event_id, [])
    activity.append(event)

gi-yt: db.get_all returns even single values inside an array, so the easiest option for me was to just run a for loop over the single value array.

harshtandiya: https://docs.frappe.io/framework/user/en/api/database#frappedbget_value

ex:
```
# as dict
task_dict = frappe.db.get_value('Task', 'TASK00002', ['subject', 'description'], as_dict=1)
task_dict.subject
task_dict.description
```

gi-yt: TIL, thanks! I'll use that syntax then

harshtandiya (approved):

PR #977, opened 2025-05-28
feat: permission handling for adding volunteers, ability to specify volunteer roles

### LINKED ISSUE #643: [FEATURE] Add ability to change member roles
### Feature description

Currently, there's no way to change the role of members in a chapter. For example, in FOSS Clubs, every member other than the lead is assigned the default role of "Core Team Member". 

### Use case

Not every member has the same role, someone could be a co-lead for example. There should be a distinguishing factor between a at a higher position and a core member.

### Benefits

_No response_

### Alternatives considered

_No response_

### Additional context

_No response_

gi-yt:
image
image
closes #643 
default role has been set to volunteer, so the role selection is totally up to the chapter lead 

image
image

### DIFF
--- a/dashboard/src/components/chapter/AddMember.vue
+++ b/dashboard/src/components/chapter/AddMember.vue
@@ -20,17 +20,53 @@
           </template>
         </Autocomplete>
       </div>
+      <div class="flex flex-col gap-2">
+        <div class="text-p-base text-gray-700">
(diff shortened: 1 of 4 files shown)


harshtandiya:
thoughts ? @rahulporuri

harshtandiya (approved):

PR #979, opened 2025-05-30
feat: show all talks a user gave on foss profile

gi-yt:
image
image

### DIFF
--- a/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
+++ b/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
@@ -8,7 +8,7 @@ from frappe.exceptions import PermissionError
 from frappe.website.website_generator import WebsiteGenerator
 
 from fossunited.api.profile import is_valid_username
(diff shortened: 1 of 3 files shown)


fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
```diff
@@ -156,7 +156,35 @@ def get_user_activity(self):
...
+                        "event_start_date",
+                        "event_name",
+                        "banner_image",
+                        "must_attend",
+                        "event_location",
+                    ],
+                    filters={"name": val.event},
+                    as_dict=1,
+                )
+                | val
```
harshtandiya, line 184: why ` | val`

gi-yt: adds talk title and the CFP route (which comes from approved_cfp_event_ids) to the dict so it can be accessed alongside the event metadata

harshtandiya (approved):
LGTM

PR #982, opened 2025-06-01
feat: proposal review notification

### LINKED ISSUE #601: [FEATURE] Proposal Review Notification
### Feature description

Request for feature: 
- Proposers should get a notification (email) whenever a review is left at their proposal.
- Reviewers should get notified when a proposal that they reviewed gets edited, for the possibility of a re-review

gi-yt:
closes #601


dashboard/src/components/reviewers/ReviewCommentBox.vue
```diff
@@ -107,6 +107,24 @@ const submitReview = () => {
...
+        },
+        auto: true,
+        onSuccess() {
+          errorMessages.value = ''
+        },
+        onError(err) {
+          errorMessages.value = err
+          toast.error('Failed to notify proposer about review', err.message)
+        },
+      })
```
harshtandiya: This should happen via server code, not like this. 

Adding a controller method to either FOSS CFP Review Doctype or the CFP Submission itself should be good IG. This is not ideal

gi-yt:
@harshtandiya i made it use after_insert hook, but the frappe email system is too fucked up for me to make it work locally. So can you test it locally and then merge if it works?

harshtandiya:
This is not working btw

gi-yt:
blocked till FOSS Event Review is made into a parent doctype

harshtandiya:
Let's close this PR then

idlip:
After some digging, found that using `db_insert` and `db_update` seems to work as intended.

Wondering if there is any edge case.

PR #984, opened 2025-06-02
fix: stop validating empty SM links

gi-yt:


dashboard/src/pages/MyProfile.vue
```diff
@@ -338,7 +338,7 @@ const updateProfileErrors = () => {
 
   for (let index = 0; index < Object.keys(socials).length; index++) {
     // error out if it doesnt match regex AND it is not empty variable
-    if ((!pattern.test(Object.values(socials)[index])) && (Object.values(socials)[index] != null)) {
+    if ((!pattern.test(Object.values(socials)[index])) && (Object.values(socials)[index] != null) && (Object.values(socials)[index] != '')) {
       errors.push('\n'+Object.keys(socials)[index] + ' is not a valid url')
     }
   }
```
harshtandiya: Wouldn't this be better and more readable

```
for (...) {
    if (!Object.values(socials)[index]){
        continue;
    }
   
    if (!pattern.test()){
        ....
    }
} 

````

gi-yt: afaik, the first if statement you gave cannot account for both null (which is given when the value is empty from the beginning) and also '' (which is given when the value is emptied by the user)

harshtandiya: image
🤔

gi-yt: @harshtandiya fixed

harshtandiya (approved):

PR #987, opened 2025-06-03
feat: og imgs for cfp submission

gi-yt:
image

### DIFF
--- a/fossunited/fossunited/doctype/foss_event_cfp_submission/foss_event_cfp_submission.py
+++ b/fossunited/fossunited/doctype/foss_event_cfp_submission/foss_event_cfp_submission.py
@@ -142,6 +142,8 @@ class FOSSEventCFPSubmission(WebsiteGenerator):
         context.review_scores = self.get_review_scores()
         context.total_reviews = len(self.reviews)
 
+        context.pagetitle, context.description, context.image = self.get_meta(context)
+
         # For Like
         context.reference_doctype = self.doctype
         context.reference_name = self.name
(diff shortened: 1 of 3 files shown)


fossunited/fossunited/doctype/foss_event_cfp_submission/foss_event_cfp_submission.py
```diff
@@ -146,6 +149,40 @@ def get_context(self, context):
...
+            + self.event_name
+            + "&speaker_designation="
+            + self.speakers[0].designation
+            + "&speaker_name="
+            + self.speakers[0].full_name
+            + "&speaker_image="
+            + self.speakers[0].photo
+            + "&event_chapter="
+            + chapter_name
+        )
```
harshtandiya: can you make seperate functions for this? looks cluttered. 

Also use `.format()` rather than + for this. That would be cleaner and more readable

harshtandiya:
Badge is broken for chapter, what's up with that

fossunited/fossunited/doctype/foss_event_cfp_submission/foss_event_cfp_submission.py
gi-yt: I added it under context function (and used + instead of format) for the event page as well, will change for both in that case

gi-yt:
> Badge is broken for chapter, what's up with that

its an svg issue, it will be like that till i find a better way of exoprting from figma to svg where the blackbox resizes according to size of text. Shouldn't affect this PR tho, since it has to be fixed on the ograph-gen server directly

harshtandiya:
can you link the code in ogen here? I would like to check it out

gi-yt:
fixed, should be good to merge @harshtandiya

gi-yt:
> can you link the code in ogen here? I would like to check it out

svgs are in the linode server that hosts vaultwarden, and conversion into png is done by an inkscape wrapper. will elaborate further in mattermost

harshtandiya:
We might need to change how things work

harshtandiya (approved):

Issue #361, opened 2024-06-03
[FEATURE] Enhancements to Event Timeline


RiyaMathew-11:
### Feature description

The event timeline page can have the following potential enhancements: 

- [X] Filter by city, chapter-type
- [ ] Search by date made more efficient

  - Current implementation is dependent on how months are rendered in the event cards
- [x] when city chapter is filtered in /events/timeline, organize upcoming/previous events monthwise

- [X] Add external events - this is referenced in #external

- [x] better logic for external events (external events shouldn't need is_published=1)

PR #988, opened 2025-06-04
feat: ability to download ics file for event

### LINKED ISSUE #641: [FEATURE] Download ICS files for all events on the Platform
### Feature description

We recently introduced the functionality to generate and download ICS files for the IndiaFOSS conference, which we should extend to all events on the platform.

This will also make it easy to maintain the public calendar as the events can simply be downloaded from the event page and uploaded to the public calendar page.

gi-yt:
image

closes #641


fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.py
```diff
@@ -219,10 +220,33 @@ def get_context(self, context):
...
+        )
+        e.location = self.event_location
+        e.organizer = self.chapter_name
+        e.url = "https://fossunited.org" + self.route
+        e.description = self.event_description
+        e.begin = self.event_start_date.replace(tzinfo=timezone(timedelta(hours=5, minutes=30)))
+        e.end = self.event_end_date.replace(tzinfo=timezone(timedelta(hours=5, minutes=30)))
+        c.events.add(e)
+
+        return c.serialize()
```
harshtandiya: make this a whitelisted method that takes the necessary arguments. This should not be a class method. 
Then call that whitelisted method using frappe.call

Name doesn't need to be so verbose. Use .format() here
Reference for previous implementation : https://github.com/fossunited/fossunited/blob/develop/dashboard/src/components/schedule/SessionCard.vue#L75

I wonder if we can just have this in js and not hit server at all.

I use the node package of ics in schedules

fossunited/chapters/doctype/foss_chapter_event/templates/foss_chapter_event.html
```diff
@@ -86,9 +86,12 @@ <h1 class="header--event-title">{{ doc.event_name }}</h1>
           <div class="event-start-date child-container--heading">
             {{ doc.event_start_date.strftime("%A, %-d %B %Y") }}
           </div>
-          <div class="event-end-date child-container--subtext">
+          <div class="child-container--subtext">
             {{ doc.event_end_date.strftime("%A, %-d %B %Y") }}
           </div>
+          <a href="" onclick="downloadCal()" class="child-container--subtext">
+            Download ICS
+          </a>
```
harshtandiya: should be a button

fossunited/chapters/doctype/foss_chapter_event/templates/foss_chapter_event.html
```diff
@@ -393,5 +396,17 @@ <h5>Live Stream</h5>
         $('#livestream_link').text('Livestreaming now!')
       }
     }
+
+    function downloadCal() {
```
harshtandiya: practice better function names

`downloadEventIcs`

pyproject.toml
```diff
@@ -11,7 +11,8 @@ dependencies = [
     "razorpay~=1.4.2",
     "PyGithub~=2.3.0",
     "Faker~=27.0.0",
-    "Pillow~=11.0.0"
+    "Pillow~=11.0.0",
+	"ics~=0.7.2"
```
harshtandiya: format

fossunited/chapters/doctype/foss_chapter_event/templates/foss_chapter_event.html
gi-yt: image
design till platform ui 3.0 is done, talked with jeswin regarding this and came to the conclusion that this is the best possible without cluttering

gi-yt:
added to /events/timeline as well
image
alongside adding it, i was able to also close point 3 on #361  (when city chapter is filtered in /events/timeline, organize upcoming/previous events monthwise)
image

gi-yt:
@harshtandiya review when you can

fossunited/fossunited/web_template/events_timeline/events_timeline.html
```diff
@@ -134,11 +159,17 @@ <h2 id="search-panel-label" class="sr-only">Search and Filter Events</h2>
         Filter the events to show only those in your city.
       </p>
     </div>
+    <button onclick="downloadEventIcs()" class="child-container--subtext ics-button">
+      Download ICS File
+      <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+        <path d="M2.66675 11.3337V12.667C2.66675 13.0206 2.80722 13.3598 3.05727 13.6098C3.30732 13.8598 3.64646 14.0003 4.00008 14.0003H12.0001C12.3537 14.0003 12.6928 13.8598 12.9429 13.6098C13.1929 13.3598 13.3334 13.0206 13.3334 12.667V11.3337M4.66675 7.33366L8.00008 10.667M8.00008 10.667L11.3334 7.33366M8.00008 10.667V2.66699" stroke="#FAFAFA" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
+      </svg>
```
harshtandiya: `<i class="ti ti-download"></i>`

We use tabler icons. See references from other places in code

fossunited/fossunited/web_template/events_timeline/events_timeline.html
```diff
@@ -178,43 +211,6 @@ <h3>{{ group }}</h3>
...
-              >
-                  {{ hackathon_card(event) }}
-              </div>
-                {% endif %}
-            {% endfor %}
-          {% endfor %}
-        {% endfor %}
-      </div>
-    </div>
-  </div>
```
harshtandiya, line 216: Why was this removed

fossunited/api/chapter.py
```diff
@@ -63,3 +66,45 @@ def check_if_event_lead(event: str) -> bool:
     )
 
     return is_lead
+
+
+@frappe.whitelist(allow_guest=True)
+def generate_ics(event_ids: str) -> str:
```
harshtandiya: event_ids should be a list of strings

fossunited/fossunited/web_template/events_timeline/events_timeline.html
```diff
@@ -249,5 +249,29 @@ <h3 id="filtered-events-label" class="sr-only">Filtered Event Results</h3>
         })
       },
     }).mount()
+
+    function downloadEventIcs() {
+      var event_ids = ''
```
harshtandiya: let / const

fossunited/fossunited/web_template/events_timeline/events_timeline.html
```diff
@@ -249,5 +249,29 @@ <h3 id="filtered-events-label" class="sr-only">Filtered Event Results</h3>
...
+      events = document.querySelectorAll('div.UpcomingFOSSEvents.event-card-holder').forEach((event) => {
+        event_ids = event_ids+','+event.id
+      })
```
harshtandiya: rather than combining strings with "," you should create an array and create a list of IDs. The method should take a list of strings

fossunited/api/chapter.py
gi-yt: Its way too complicated to send a list using frappe.call.

fossunited/fossunited/web_template/events_timeline/events_timeline.html
gi-yt: Again, need comma separated since frappe.call doesn't work properly with list arguments

fossunited/fossunited/web_template/events_timeline/events_timeline.html
gi-yt: The content from this block was moved back to the original block, since it was only used if search text != ""

fossunited/api/chapter.py
harshtandiya: how come? We can pair program an sort this out if needed. 

But this is not a good pattern, converting list to commas seperated string, only to convert it to list again on server side makes no sense

harshtandiya:
image
Nitpicky, but the design can be better. Also the text should be "Add to Calendar"

fossunited/fossunited/web_template/events_timeline/events_timeline.html
```diff
@@ -1,6 +1,11 @@
 {% from "fossunited/templates/macros/event_card.html" import event_card %}
 {% from "fossunited/templates/macros/hackathon_card.html" import hackathon_card %}
 
+<link
+  rel="stylesheet"
+  href="https://cdn.jsdelivr.net/npm/@tabler/icons-webfont@latest/tabler-icons.min.css"
+/>
+
```
harshtandiya, line 8: I don't think you need to do this, afaik, tabler is a node module. You can directly use it

gi-yt: we need it since events timeline doesnt use foss_base.html template

gi-yt: same thing is done in other files too, such as fossunited/fossunited/web_template/home_page___events/home_page___events.html

harshtandiya (approved):

PR #1013, opened 2025-06-14
fix: give out error properly if image isnt png/jpeg

gi-yt:


gi-yt:
@gunjjoshi this should fix your issue

harshtandiya (approved):

PR #1032, opened 2025-06-22
fix: show speakers tab to everyone

### LINKED ISSUE #1031: Speakers tab not visible without login on event page
When visiting the following event page without being logged in:
🔗 https://fossunited.org/c/chennai/2025/june

The "Speakers" tab is not visible in the navigation bar.

This affects users who want to view the event details and speaker lineup before logging in or registering. The tab should ideally be accessible to all visitors, even if they are not logged in.

# Steps to Reproduce:

Open the event link in an incognito/private window (or log out).

Notice that the “Speakers” tab is missing from the top navigation.

# Expected Behaviour:
The "Speakers" tab should be visible and accessible without requiring login.

# Actual Behavior:
The tab is hidden unless the user is logged in.

gi-yt:
closes #1031


harshtandiya (approved):

rahulporuri:
I'm not sure why this was merged in the first place :sweat:

Issue #1045, opened 2025-06-24
Event page doesn't respect "Show Speakers" checkbox


rahulporuri:
Ref https://fossunited.org/c/indiafoss/maintainer-summit-2025

"Show speakers" checkbox is unchecked for the "Maintainer Summit 2025" but the tab is still displayed on the page

PR #1046, opened 2025-06-24
Revert "fix: show speakers tab to everyone"

harshtandiya:
Reverts fossunited/fossunited#1032


harshtandiya:
cc: @gi-yt 

We should rather show the checkbox for organizers to enable or disable it via dashboard 
#1045

PR #1051, opened 2025-06-26
feat: fossprofile og images!

gi-yt:
image

images will work on prod

## Summary by CodeRabbit

* **New Features**
  * User profile pages now display improved metadata, including custom page titles, descriptions, and images for better sharing and visibility.
  * Private profiles show limited metadata to protect user privacy.

### DIFF
--- a/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
+++ b/fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
@@ -329,14 +329,14 @@ class FOSSUserProfile(WebsiteGenerator):
         if self.about:
             desc_short = textwrap.shorten(re.sub(r"<.*?>", "", self.about), width=150)
 
-        description = "{self.full_name} is a FOSS United Community Member. {desc_short}".format(  # noqa: E501
+        description = "{self.full_name} is a Community Member. {desc_short}".format(  # noqa: E501
             self=self,
             desc_short=desc_short,
         )
 
         og_url = frappe.db.get_single_value("Ograph Settings", "ograph_url")
 
-        image = "{og_url}/gen/fossprofile?username={self.username}&full_name={self.full_name}&designation={self.bio}&image={self.profile_photo}".format(  # noqa: E501
+        image = "{og_url}/gen/profile?username={self.username}&full_name={self.full_name}&designation={self.bio}&image={self.profile_photo}".format(  # noqa: E501
             self=self, og_url=og_url
         )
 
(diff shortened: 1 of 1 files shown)


harshtandiya:
@gi-yt you don't have to address everything by coderabbit but make changes where you can as per suggestions

harshtandiya:
Also, in case of no images, we should use default image imo. You can find it in the codebase

gi-yt:
fetching the default image will be too heavy bandwidth wise without much benefit imo

rahulporuri:
> fetching the default image will be too heavy bandwidth wise without much benefit IMO

agreed. Unless ograph-gen has a different template file that contains the default image.

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
```diff
@@ -313,8 +314,34 @@ def get_context(self, context):
...
+            return self.username, "Private Profile", ""
+
+        # eg. Arya | arya_k
+        pagetitle = self.full_name + " | " + self.username
+
+        desc_short = ""
+        if self.about:
+            desc_short = textwrap.shorten(re.sub(r"<.*?>", "", self.about), width=150)
+
+        description = "{self.full_name} is a FOSS United Community Member. {desc_short}".format(  # noqa: E501
```
rahulporuri: The URL will already say fossunited.org, so any specific reason why we want to add the "... is a FOSS United Community member" - it's extra text, and I'm not sure what additional information it conveys?

fossunited/foss_profiles/doctype/foss_user_profile/foss_user_profile.py
```diff
@@ -313,8 +314,34 @@ def get_context(self, context):
...
+            self=self,
+            desc_short=desc_short,
+        )
+
+        og_url = frappe.db.get_single_value("Ograph Settings", "ograph_url")
+
+        image = "{og_url}/gen/fossprofile?username={self.username}&full_name={self.full_name}&designation={self.bio}&image={self.profile_photo}".format(  # noqa: E501
```
rahulporuri: this needs to happen on ograph-gen but I would prefer to not spread the "FOSS" prefixes everywhere - how about just `/gen/profile?...` instead of `/gen/fossprofile?...`

harshtandiya:
If we can shift the templating from SVG, to something more standard like jinja / html, we can define the style easily and place base64 values for these images directly. That should take care of things no?

harshtandiya:
Maybe it can be still possible?

gi-yt:
image
fixed

harshtandiya:
fix linters pls

harshtandiya:
nvm, not your fault

harshtandiya (approved):

PR #1080, opened 2025-07-27
Auto close jobs after threshold duration

### LINKED ISSUE #1066: Auto-expire jobs after a set duration
Instead of manually marking old job postings as "Expired", it'll be good to autoclose them using a cron job e.g. once a month, mark every job posting that is older than 3 months as "Expired".

We can't do this silently either - the cron job should also trigger an email to the relevant poster + cc the foundation, in case the job posting could be updated appropriately.

idlip:
+ Set frappe schedules to auto expire jobs older than 90 days (3 months) that are Active (Approved) till now.
+ Scheduled to run monthly to check.
+ Sends email to Company and CC fossunited foundation for informing about the action.

Closes #1066

## Summary by CodeRabbit

* **New Features**
  * Job postings approved but inactive for over 90 days are now automatically marked as expired monthly.
  * Job posters receive email notifications when their job posting expires due to inactivity.

### DIFF
--- a/fossunited/scheduled_tasks.py
+++ b/fossunited/scheduled_tasks.py
@@ -1,8 +1,8 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 
 import frappe
 
-from fossunited.doctype_ids import EVENT
+from fossunited.doctype_ids import EVENT, JOB, JOB_STATUS_APPROVED, JOB_STATUS_EXPIRED
 
 
 def conclude_events():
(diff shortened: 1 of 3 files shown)


fossunited/scheduled_tasks.py
```diff
@@ -27,3 +27,57 @@ def conclude_events():
...
+
+
+def update_past_job_status():
+    """
+    Updates the job status to Expired, if they are Approved (active) and
+    were last modified more than 3 months ago (90 days).
+    """
+    cutoff_date = datetime.now() - timedelta(days=90)
+    past_active_jobs = frappe.get_all(
+        "Job Board",
```
rahulporuri: it'll be good to add this constant to `fossunited.doctype_ids` e.g. `JOB`

fossunited/scheduled_tasks.py
```diff
@@ -27,3 +27,57 @@ def conclude_events():
...
+        filters={"status": "Approved", "modified": ["<", cutoff_date]},
```
rahulporuri: `"Approved"` should similarly be a constant in this python module e.g. `JOB_STATUS_APPROVED`

fossunited/scheduled_tasks.py
```diff
@@ -27,3 +27,57 @@ def conclude_events():
...
+        fields=["name"],
+    )
+
+    for jobs in past_active_jobs:
+        try:
+            # Get the actual document object
+            job_doc = frappe.get_doc("Job Board", jobs.name)
+            job_doc.status = "Expired"
```
rahulporuri: similar comments as above

- use the previously defined `JOB` constant
- create and use a new `JOB_STATUS_EXPIRED` constant

fossunited/scheduled_tasks.py
```diff
@@ -27,3 +27,57 @@ def conclude_events():
...
+            job_doc.save(ignore_permissions=True)
+            frappe.db.commit()
+
+            # Send notification email after successful update
+            if job_doc.mail:
+                cc_email = ["foundation@fossunited.org", "developers@fossunited.org"]
```
rahulporuri: let's remove `foundation@fossunited.org` from the list here

fossunited/scheduled_tasks.py
```diff
@@ -27,3 +27,57 @@ def conclude_events():
...
+                    recipients=[job_doc.mail],
+                    cc=cc_email,
+                    subject=subject,
+                    message=message,
+                )
+        except Exception as e:
+            frappe.log_error(
+                frappe.get_traceback(),
+                f"""Error while updating status of past jobs as Expired through scheduler
+                - ID:{jobs.name}\nError:{str(e)}""",
```
rahulporuri: why triple quotes? why not just simple strings `"..."`?

rahulporuri (commented):
This looks correct, but were you able to test this locally? I've made a few cosmetic change requests but the big question is whether or not this has been tested.

fossunited/scheduled_tasks.py
idlip: The linter complains about long line, so I thought to get it at new line. The indentation gets more spaces.

or would it be fine to minimize text content?

rahulporuri: I'm okay with smaller text - `"Error updating job status"`

fossunited/scheduled_tasks.py
idlip: I was not able to test it. I think I've to setup proper outgoing email to test this?
So locally should I set working smtp email? gmail?

Please let me know if I'm correct or is there other ways to test.

rahulporuri: let's not worry about the email - I want to ensure that the cron job works as expected e.g. change it to monthly job to an hourly job and create a job with an old enough day. what do you think?

idlip: Okay I confirm that the scheduler works as intended.

image

I had some trouble getting scheduler in frappe. It was not auto running it, but simply uninstall and install of fossunited app worked.

I manually tested it via console as well. The job status changes were reflected.

rahulporuri (approved):
Thanks for testing this @idlip , merging this now

fossunited/scheduled_tasks.py
rahulporuri: thanks @idlip

PR #1088, opened 2025-08-10
show schedule pages with correct route

idlip:
- idk why c/ was pre-pended here

Issue opened by @ansharora28

Currently there is no `schedule` page shown for any chapter events.

Eg: https://fossunited.org/dashboard/event/6ujtqen8oh/schedule

The page gives out 404 not found:
https://fossunited.org/schedule/c/bengaluru/aug-2025

Actually the page exists here:
https://fossunited.org/dashboard/schedule/bengaluru/aug-2025

- This was an minor API issue.
Which pre-pended `c/` to the route page.
But this conflicts or is taken care over at `hooks.py` via redirect.

I might be wrong with reasoning... Happy if its works ;)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed route handling when retrieving event data: routes are now used exactly as provided (no automatic prefixing), preventing mismatches and ensuring more consistent, predictable lookups. This reduces unexpected behavior when accessing event information and improves reliability across different route formats.

PR #1089, opened 2025-08-11
revert and actual fix the schedule page redirect

idlip:
- actually it was my local setup issue before, but now fixed the actual issue on production side

reverting and fixes #1088

## Summary by CodeRabbit

- Bug Fixes
  - Corrected the “Go to Schedule Page” link to use the <route>/schedule path, ensuring consistent navigation while still opening in a new tab.
  - Improved event route handling to accept routes with or without the “c/” prefix, reducing broken links and improving page resolution.

### DIFF
--- a/dashboard/src/pages/EventSchedule.vue
+++ b/dashboard/src/pages/EventSchedule.vue
@@ -176,7 +176,7 @@ const showModifyScheduleItemDrawer = ref(false)
           variant="ghost"
           label="Go to Schedule Page"
           icon-right="arrow-up-right"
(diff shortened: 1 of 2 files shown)


idlip:
Ok after above coderabbit suggestion.
I think this can be solved in 2+ ways.

The one merged might work fine...

It can also be solved via
`@click="redirectRoute(`dashboard/schedule/${event.doc?.route}`, '_blank')"`

since the page is rendered via base dashboard.

But I think, its safe to ignore coderabbit lint above.

PR #1102, opened 2025-08-20
close rsvp and cfp form when event ends

### LINKED ISSUE #1096: Auto close RSVP forms after event ends
Right now RSVP form is unpublished/closed when only max submissions are reached.

Ideally it should close when the event ends.

---

One way to solve is to trigger some form when event is concluded, close the rsvp form as well.

or apply scheduler/cron job to check every week for concluded events.

idlip:
- helps to hide forms and avoid manual closure by team

closes #1096

Unchecks (close) both RSVP and CFP form after event ends.

## Summary by CodeRabbit

* **Bug Fixes**
  * Scheduled task now processes both Live and Cancelled events, marking them Concluded or Cancelled, clearing RSVP/CFP visibility, and updating linked RSVP forms to be unpublished and CFPs to be closed.
* **Tests**
  * Added tests and helpers validating event conclusion behavior and RSVP/CFP state transitions across Live, Cancelled, and Concluded scenarios (with controlled current-date patching).

### DIFF
--- a/fossunited/scheduled_tasks.py
+++ b/fossunited/scheduled_tasks.py
@@ -18,14 +18,20 @@ def conclude_events():
     """
     events = frappe.db.get_all(
         EVENT,
-        {"status": "Live", "event_end_date": ["<", datetime.today()]},
+        {
+            "status": ["in", ["Live", "Cancelled"]],
+            "event_end_date": ["<", datetime.today()],
+        },
         ["name", "status", "event_end_date", "event_start_date"],
         page_length=999,
     )
 
     for event in events:
         doc = frappe.get_doc(EVENT, event.name)
-        doc.status = "Concluded"
+        if doc.status == "Live":
+            doc.status = "Concluded"
+        else:
+            doc.status = "Cancelled"
         doc.show_rsvp = 0
         doc.show_cfp = 0
         past_rsvp = frappe.get_doc(EVENT_RSVP, {"event_name": doc.event_name})
--- a/fossunited/scheduled_tasks.py
+++ b/fossunited/scheduled_tasks.py
@@ -19,6 +19,8 @@ def conclude_events():
     for event in events:
         doc = frappe.get_doc(EVENT, event.name)
         doc.status = "Concluded"
+        doc.show_rsvp = 0
+        doc.show_cfp = 0
         try:
             doc.save(ignore_permissions=True)
         except Exception as e:
--- a/fossunited/scheduled_tasks.py
+++ b/fossunited/scheduled_tasks.py
@@ -2,7 +2,14 @@ from datetime import datetime, timedelta
 
 import frappe
 
-from fossunited.doctype_ids import EVENT, JOB, JOB_STATUS_APPROVED, JOB_STATUS_EXPIRED
+from fossunited.doctype_ids import (
+    EVENT,
+    EVENT_CFP,
+    EVENT_RSVP,
+    JOB,
+    JOB_STATUS_APPROVED,
+    JOB_STATUS_EXPIRED,
+)
 
 
 def conclude_events():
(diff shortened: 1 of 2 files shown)


fossunited/scheduled_tasks.py
```diff
@@ -19,6 +19,8 @@ def conclude_events():
     for event in events:
         doc = frappe.get_doc(EVENT, event.name)
         doc.status = "Concluded"
+        doc.show_rsvp = 0
+        doc.show_cfp = 0
```
rahulporuri, line 37: this will simply hide the rsvp and cfp forms linked with the event - but not "unpublish" or "close" them

for that, we will need to pull the associated CFP and RSVP doctypes and update the `status` on CFP to `Closed` and `is_published` on RSVP to `0`. Ref `FOSSEventCFP` and `FOSSEventRSVP` doctypes

idlip: Ohh point..
Sorry for missing the flow here.

Will update!

idlip: Pushed the fix.

idlip:
@rahulporuri Pushed the fix.

Now it fetches RSVP and CFP doc info for and sets CFP status to closed and rsvp to be unpublished.

idlip:
Hold! I will investigate unit case failing!

idlip:
Latest change:

- Added unit tests

Made concluding event also check for cancelled and close RSVP and CFP is they are open (just to be safe)

- Added unit test to include both forms based on 3 events
  - event1 was live and ended so test checks to be concluded and if both forms are closed.
  - event2 was live but cancelled, so again we expect same to be closed
  - event3 is live and both should be open/published.

Unit test pass. The scheduler should close event, and both forms as expected.

rahulporuri (approved):
this looks good - lets merge ~and check tomorrow to see if some of the open RSVPs/CFPs got closed.~

sorry - I think you'll need to manually close/unpublish some of the current RSVPs/CFPs because this script won't close them

fossunited/scheduled_tasks.py
```diff
@@ -11,16 +18,30 @@ def conclude_events():
...
         page_length=999,
     )
 
     for event in events:
         doc = frappe.get_doc(EVENT, event.name)
-        doc.status = "Concluded"
+        if doc.status == "Live":
+            doc.status = "Concluded"
+        else:
+            doc.status = "Cancelled"
```
rahulporuri, line 34: why do we need this else?

idlip: From doctypes, we grab all events that are ended but are [Live & Cancelled]
Actually we close only for live events.

I added `else` for cancelled events cause there might be any RSVP or CFP forms still open.
Although ideally I dont think cancelled would have them open. But just a safer check to turn them if found.

Only special case this applies is, an event is made which ends next week with both forms open/live, then it got cancelled due to some reason. So this scheduler should act as safer check to close them if so found

fossunited/scheduled_tasks.py
```diff
@@ -11,16 +18,30 @@ def conclude_events():
...
+        doc.show_cfp = 0
+        past_rsvp = frappe.get_doc(EVENT_RSVP, {"event_name": doc.event_name})
+        past_rsvp.is_published = 0
+        past_cfp = frappe.get_doc(EVENT_CFP, {"event_name": doc.event_name})
+        past_cfp.status = "Closed"
         try:
             doc.save(ignore_permissions=True)
+            past_rsvp.save(ignore_permissions=True)
+            past_cfp.save(ignore_permissions=True)
         except Exception as e:
```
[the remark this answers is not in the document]
idlip: makes sense. Added it!

idlip:
hopefully it should work.

Logic seems fine, if I missed something, we will know ahead :)

Issue #1151, opened 2025-09-11
Support more download format for Event schedule page


idlip:
We only added calendar download, and that itself is hit or miss for many

Since it clutters calendar for them.

Many people just expect some text or simple format to note down and follow.

PR #1152, opened 2025-09-11
Download schedule in various formats

idlip:
## 🚦 Status

- [x] Ready for review

---

## 🔗 Related Issue

Closes / Addresses: #1151

---

## ❗ Problem

More than ICS, txt csv are more preferred for user to filter on their own
and again, plain text are easy to search via grep, Ctrl-f and more...

---

## ✅ Solution

Simply try to format for various file formats. Since its easy as plain text

---

- [x] Tested locally

---

## 📝 Changelog

<commit>: 

---

## 🖼️ Visualization

Here is demo video of it.

https://github.com/user-attachments/assets/c1b11ac1-a292-4ed7-9455-7c36963ef125

---

## 🔮 Further Ahead

Improve PDF and make it more designed and beautiful?

Lot of content can be added. I believe `typst` can be best for pdf...

## Summary by CodeRabbit

* **New Features**
  * Added a “Download Schedule” flow: choose format (ICS, CSV, TXT, PDF, JSON, Markdown/Org), select days and halls, then open/download filtered exports; filenames reflect chosen filters.
  * Server-side export endpoint supports multiple formats, timezone-aware times, sanitized CSV cells, and returns appropriate empty outputs when no sessions match.

* **Refactor**
  * Replaced the previous calendar-export UI with the new Schedule Download experience while preserving layout and props.


idlip:
Merging this to see it in action.

All download format work as far I tested.

This coderabbit will be suggesting more code to handle things safely, although we are confident that the doctype filled will have details as we know.
Tbh so annoying that we have to think of "what if" we dont have that add, this or in that datatype...

PR #1183, opened 2025-09-30
UX: show disabled + expired tickets for users to refer

idlip:
- We will make inactive/expired tickets as non-clickable so it acts as only reference to stay transparency and avoid mis-perception

- Referred via mail
  > Re: Request for Extension of Regular Ticket Price end date.

Copying the same mail content below:

> This is a UX issue Dilip  - the tiers shouldn't disappear from the tickets page when they are sold out, but
instead be unselectable and have a "Sold out" on "X date" text.
> We don't want to give off the perception that we are dramatically increasing the prices, especially to previous
attendees.

CC - @rahulporuri (for any suggesstion further)

TLDR:

- Even if tickets are disabled in our end, it appears as closed
- If ticket closing date exceeds, it shows as closed for that card tier
- Closed tickets are not clickable

Here is demo for visualization:

https://github.com/user-attachments/assets/7cc05fb0-3d5b-499d-890c-9c87abe2b2dd

## Summary by CodeRabbit

* **New Features**
  * Per-tier availability state drives UI gating; first active tier auto-selected on load. Pay Now disabled until a valid tier is chosen with inline prompt.

* **Bug Fixes**
  * Prevents selecting inactive/expired tiers and invalidates selections when availability or seat counts change. Summary/sidebar/mobile handle missing or inactive tiers.

* **Style**
  * Updated tier cards, badges, outlines, checkmarks, and price presentation to consistently reflect active/disabled states.

### DIFF
--- a/dashboard/src/pages/BuyTickets.vue
+++ b/dashboard/src/pages/BuyTickets.vue
@@ -366,6 +366,7 @@
         @click="showAdditionalDetails = !showAdditionalDetails"
       />
     </div>
+
     <div v-if="showAdditionalDetails" class="w-full mt-2 space-y-1">
       <div class="grid grid-cols-3 gap-2">
         <p>{{ checkoutInfo.tier?.title }} Ticket</p>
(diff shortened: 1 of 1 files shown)


idlip:
Merging this as it looks fine for now.
Might step in later if found any issues..!

PR #1194, opened 2025-10-08
docs: add cname for docs deploy

idlip:
maybe it fixes re-direction..

if not revert!


idlip:
BINGO! Fixed.

Hopefully it won't happen ever again!

Issue #1196, opened 2025-10-08
Events Chapter filter can be improved


HarshPatel5940:
1. What is "foss x policy" and "oasis" considered as an city chapter? somethings wrong
2. The dropdown is default dropdown, no style, no type filtering, no multi select filtering... its visually bad and functionally hard..

<img width="3024" height="1964" alt="Image" src="https://github.com/user-attachments/assets/537fb62c-d529-480b-9b93-9870448f8e5e" />

HarshPatel5940:
If there are no upcoming events for an month, just dont show the month name

HarshPatel5940:
> If there are no upcoming events for an month, just dont show the month name

@idlip should this be an seperate issue?

idlip:
Hi,

With #1198 I'm closing this.

I only made the dropdown to work for club as well (previously it did not)

native select option does not do multi select and search both.

We might need to depend on another dependency.

Maybe next time when cards clutter more. Please feel free to contribute !

PR #1198, opened 2025-10-09
Ux timeline

idlip:
## 🚦 Status

- [x] Draft
- [ ] Ready for review

---

## 🔗 Related Issue

Closes / Addresses: #1196

---

## ❗ Problem

- City filter only applied for City community names
- some UX can be improved (as harsh patel says -> Design language)

---

## ✅ Solution

- Apply city filter to all events based on their chapter city

Note: We've created some city names to some Chapters previously for convenience to fill in arbitrary (we should resolve this and strictly adhere to existing city)

---

- [x] Tested locally
- [ ] Theme the select dropdown, don't let browser natively handle it. (We gonna force it?)

---

## 📝 Changelog

<commit>: 

---

## 🖼️ Visualization

---

## Summary by CodeRabbit

* **New Features**
  * Chapter form auto-fills city, state and chapter name when an institution is selected.
  * Events timeline: added city selector, event-type filtering, and improved city-based visibility logic; search now includes city and event type.
  * City list exposed to templates for populating city select controls.

* **Bug Fixes**
  * Improved error handling for institute lookups with user-facing error messages.

* **Style**
  * Shorter truncation for event and chapter text for a cleaner layout.

### DIFF
--- a/fossunited/chapters/doctype/foss_chapter/foss_chapter.js
+++ b/fossunited/chapters/doctype/foss_chapter/foss_chapter.js
@@ -17,11 +17,20 @@ frappe.ui.form.on('FOSS Chapter', {
 frappe.ui.form.on('FOSS Chapter', {
   institution_name: function (frm) {
     if (frm.doc.institution_name) {
-      frappe.db.get_doc('Institute', frm.doc.institution_name).then(function (inst) {
-        frm.set_value('city', inst.city)
-        frm.set_value('state', inst.state)
-        frm.set_value('chapter_name', inst.name)
-      })
+      frappe.db
+        .get_doc('Institute', frm.doc.institution_name)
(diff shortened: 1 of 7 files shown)


idlip:
Re-visiting again.

Only change is:

- Made city filter apply for Clubs as well

   Simply we will fetch city list (doctype) and map those to event.chapter data
   
With that I'll merge this.

I'm not adding select option tweaks, anyone can contribute if so. Natively we cannot have multi-select and search with boostrap. Might need to use another dependency.

rest all changes are:

1. Foss_chapter: For foss club if institute is set, made city and state to inherit from it.
   Since its makes sense to link and map those from there.
   So js frontend and backend does it
   
2. event_timeline.html file was formatted using djinja tool (jinja formatter)

PR #1245, opened 2025-10-31
init: PWA support for dashboard via vite plugin

idlip:
- @nammahari has suggested to have PWA support for dashboard so users can have site installed as app
(technically its browser running the site)

- We can easily get PWA working via vite plugin, but now its a question of do we need to add dependency on dashboard. 
- the worker is quite massive package [refer here](https://vite-pwa-org.netlify.app/workbox/)

Although the debate can be dashboard has not much to offer for user to keep opening (as of now) compared to social media.

Its just a convenience.

Note: I'm opening this PR for reference and show of possibility for future as things might improve

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -36,6 +36,7 @@
     "postcss": "^8.4.31",
     "prettier": "3.2.5",
     "tailwindcss": "^3.2.7",
-    "vite": "^5.4.21"
+    "vite": "^5.4.21",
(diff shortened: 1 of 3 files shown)


idlip:
Try out frappe's mobileNav

- https://ui.frappe.io/docs/components/mobilenav
- https://ui.frappe.io/docs/components/mobileshell

They can be used to make pretty looking dashboard

PR #1259, opened 2025-11-04
backend: import free code doctype into codebase to make it generic

idlip:
BACKEND method for #999 

Changed "IndiaFOSS Free Ticket Code" -> "Event Free Ticket Code" 
same for Free ticket application

- [X] Applied the server side script in doctype before_save itself
  Only application is taken for valid ticket code as per event

- [X] Added ticket transfer protection to not transfer "Free pass" tickets (they can contact us)

NOTE:
In fossu desk I've renamed the doctype to make migration easier.

There is no code using prev code, so everything must be good.

## Summary by CodeRabbit

* **New Features**
  * Free ticket application form for requesting complimentary event tickets
  * Coupon-based free ticket codes for organizers to issue and track usage

* **Improvements**
  * Automatic ticket creation when a valid free-ticket application is submitted
  * Transfers blocked for restricted "Free Pass" tickets; transfers allowed only for active (Live) events

* **Tests**
  * Comprehensive tests covering application flow, coupon validation, usage counting, state transitions, and error cases

PR #1261, opened 2025-11-05
doctype: only permit team to modify their free ticket codes

idlip:
- for #1259 

- To let only chapter team or desk users to modify free ticket codes.

- we make this to avoid modifying other free ticket codes 

## Summary by CodeRabbit

* **New Features**
  * Permission-based controls for modifying or deleting event free ticket codes — only administrators and authorized core team members can make changes; unauthorized attempts are blocked with an error.
* **Bug Fixes**
  * Added validation to prevent using already-used or fully redeemed free ticket codes, stopping duplicate or over-limit redemptions.


idlip:
TLDR;

Just some permission to check so no other can cross modify/delete other ticket code
In case someone tries to hack us out!

Also
- Set emailing message in desk to be dynamic, so it applies as generic to all events

PR #1309, opened 2025-12-04
backend(cfp): notify chapter member if approved proposal has been withdrawn

idlip:
- Event members would like to get notification if approved proposer has withdrawn their proposal This helps them to keep track and get notified formally.

- We already allow them to download CSV for all proposals, but still not every time they would visit and screen.

## Summary by CodeRabbit

* **New Features**
  * Notify chapter core team members by email when an approved proposal is withdrawn.
  * New API to fetch chapter member emails and exposed to templating/Jinja.

* **Improvements**
  * Refined withdrawal handling to set appropriate submission statuses and trigger the new notification flow.
  * Added a distinct "Core Team" role identifier for chapter membership.

### DIFF
--- a/fossunited/api/chapter.py
+++ b/fossunited/api/chapter.py
@@ -10,6 +10,7 @@ from ics import Calendar, Event
 from fossunited.doctype_ids import (
     CHAPTER,
     CHAPTER_MEMBER,
+    CORE_TEAM,
     EVENT,
     EVENT_CHECKIN,
     EVENT_VOLUNTEER,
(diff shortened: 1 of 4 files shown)


idlip:
We will better handle sending mails via notification (desk)

idlip:
Nope, we cannot do via desk.

And its better to handle in codebase to stay transparent and if anyone want to add suggestion.

- Note: I tried to do this email via frappe desk notification
  But there was no easy way to trigger this via value_change or method before_save

method before_save fails cause i think our codebase runs before which spoils the condition.
  We had to set cond like: doc.is_withdrawn and doc.status == "Approved"

But value change does not trigger. trigger method: validate and before_validate does work, but they
don't continue our frappe controller (idk why)


And as much I explored the notification doctype code all it does is send mail and add some feature like community (which shows up in activity) and so.

But we can easily manage sending mails in our codebase as well by making a file for all message content and controlling via frappe controller.

PR #1312, opened 2025-12-05
feat: proposal review notification

idlip:
- #601 (taking prev work based on #982)

Credits: Aryak

Battles with child table to handle controller, after exploring frappe.model.document file we have _doc_before_save method and since reviews are actually less given, we can simply evaluate those by loop.

- Tested, this works by sending email!

To test can replace email with:

frappe.log_error(
    title="A review on the proposal {self.name}",
    message=build_review_message(self, review, "new"),
)

Will add screenshot later.

## Summary by CodeRabbit

* **New Features**
  * Proposers and listed speakers receive automated emails when a new review is added or reviewer remarks are changed; messages include prior values and clearer subject prefixes and link to the proposal.
* **Bug Fixes**
  * Notifications reliably trigger on save so proposers stay informed of review and approval changes.
* **Tests**
  * Expanded coverage: withdrawals, review notifications (new/remarks-changed), score calculations, permission checks, and email assertion helpers.

### DIFF
--- a/fossunited/fossunited/doctype/foss_event_cfp_submission/foss_event_cfp_submission.py
+++ b/fossunited/fossunited/doctype/foss_event_cfp_submission/foss_event_cfp_submission.py
@@ -105,6 +105,7 @@ class FOSSEventCFPSubmission(WebsiteGenerator):
         self.validate_session_type_permissions()
         if self.has_value_changed("subscribe_chapter_mailing"):
             self.handle_email_group("CFP Proposers")
+        self.notify_proposer_on_review()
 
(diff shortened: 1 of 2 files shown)


idlip:
Ah, one issue is to which email to send these...

Ideally proposer themselves must fill and apply proposal via fossunited account (so email gets record)

But for IF25 events, rahul and staff had filled/created on their behalf, so main email is collected on who submitted (submitted_by: linked DF)

So now for robustness, we might send to=self.email 

and cc to all speakers email listed

PR #1332, opened 2025-12-12
api(tickets): refactor ticket insight function to be simple

idlip:
- following the suite of #1327 

- Backend function gives out simple flat list of ticket data 
- Nothing major, just some better code

## Updates

* Event custom fields now dynamically appear in the ticket list, enabling greater flexibility in per-event customization and data collection
* T-shirt field presentation has been refined for improved usability
* Ticket data retrieval and system performance have been optimized


idlip:
tested again. works as expected

PR #1401, opened 2026-02-03
api: implement team member check in project creation (fixes unauthori…

ni5arga:
This PR primarily patches an access control vulnerability in the `create_project` API endpoint. Previously, any authenticated user could create a project for any team ID. This change creates a mandatory check to ensure the `frappe.session.user` is an active member of the specified team before allowing project creation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Project creation now requires team membership: only members of a team can create projects for that team; unauthorized attempts are blocked.

### DIFF
--- a/fossunited/api/hackathon.py
+++ b/fossunited/api/hackathon.py
@@ -219,6 +219,25 @@ def create_project(hackathon: str, team: str, project: dict) -> dict:
     Returns:
         dict: Project document as a dictionary
     """
+    team_doc = frappe.get_doc(HACKATHON_TEAM, team)
+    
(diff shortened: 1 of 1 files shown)


idlip:
LGTM.

Thanks for quick fix.

Can you please add a rate limiter as well.

We've so many api endpoints for hackathon, I'll take time to think if we can use frappe-ui method to do this via frontend.

If not I'll be making decorators to easily apply permission based calls for all functions.
Same will apply for chapter/events fetching api

idlip:
Thank you!

Issue #1278, opened 2025-11-17
Cannot add experiences, education, or projects on the profile page


agriyakhetarpal:
Hi! On `https://fossunited.org/u/<username>`, it appears that the buttons to add experiences, education, and/or projects are no-ops. I experienced this when I tried to update these sections on my profile to experiment with.

If I navigate to the dashboard by clicking on the "Edit profile" button, I cannot see an option to update my profile beyond "Basic details" and "Social links", so it looks like the above link is the only place where one should be able to edit these fields.

idlip:
Hi agriya.

Nice to see you active here!

Yea I had it hold based on #529 

But our designed (@jeswinjosu ) informed that he has plans to improve and make it usable.
So yea, thanks for bringing this to attention, we can push it.

there were other users requesting the same too!

agriyakhetarpal:
Thanks for the update, @idlip! I'm glad I could bring this to your attention. I think the idea in #529 is to remove them, which probably makes sense as a temporary solution while this is being redesigned, so that broken/incomplete functionality is not exposed to users.

idlip:
Okay, hiding it for now.

Since until now nobody has anything put up there.
Should open up discussion with @rahulporuri @jeswinjosu

ni5arga:
@agriyakhetarpal you should be able to add projects to your FOSS United profile now, added the option to add projects to user profile in PR #1459  :)

agriyakhetarpal:
> [@agriyakhetarpal](https://github.com/agriyakhetarpal) you should be able to add projects to your FOSS United profile now, added the option to add projects to user profile in PR [#1459](https://github.com/fossunited/fossunited/pull/1459) :)

@ni5arga amazing! I'd love it if it could pick my projects up from [my Forklore profile too](https://forklore.in/maintainers/agriyakhetarpal). See #1279 for more! :D

PR #1459, opened 2026-03-05
feat(profile): add projects section to user profile

ni5arga:
## Status

- [x] Ready for review

## Related Issue

Closes / Addresses: #1278

- [x] Tested locally

## Changelog

- Added a Projects section with full CRUD: view all projects, inline edit form per project, add new project form, delete with confirmation

- Initially added 4 whitelisted API endpoints (get_profile_projects, add_profile_project, update_profile_project, delete_profile_project), then removed them all in a subsequent refactor since createDocumentResource handles persistence directly via frappe.client.save

### Screenshots

image

image

### DIFF
--- a/dashboard/src/pages/MyProfile.vue
+++ b/dashboard/src/pages/MyProfile.vue
@@ -620,6 +620,7 @@ const validateProject = (p) => {
 
 const openAddProject = () => {
   Object.assign(newProject, emptyProject())
+  editingProject.value = null
   projectErrors.value = ''
   showAddProject.value = true
 }
@@ -643,6 +644,7 @@ const handleAddProject = () => {
 }
 
 const startEditProject = (project) => {
+  showAddProject.value = false
   editingProject.value = { ...project }
   projectErrors.value = ''
 }
--- a/dashboard/src/pages/MyProfile.vue
+++ b/dashboard/src/pages/MyProfile.vue
@@ -187,22 +187,17 @@
             <FormControl v-model="profileDoc.doc.medium" type="url" label="Medium" />
             <FormControl v-model="profileDoc.doc.mastodon" type="url" label="Mastodon" />
             <FormControl v-model="profileDoc.doc.bluesky" type="url" label="Bluesky" />
-            <ErrorMessage class="col-span-2" :message="updateErrors" />
-            <div class="hidden md:block"></div>
-            <div class="flex justify-end">
-              <Button
-                :variant="'solid'"
-                :size="'md'"
-                :theme="'green'"
-                label="Save"
-                class="w-full md:w-2/3"
-                @click="handleUpdateProfile()"
-              />
-            </div>
-
             <!-- Projects -->
-            <div class="col-span-2 py-1 border-b mt-2">
+            <div class="col-span-2 py-1 border-b mt-2 flex items-center justify-between">
               <h4 class="text-md font-medium uppercase">Projects</h4>
+              <Button
+                v-if="!showAddProject"
+                variant="outline"
+                size="sm"
+                icon="plus"
+                label="Add Project"
+                @click="openAddProject()"
+              />
             </div>
             <div class="col-span-2 text-sm text-ink-gray-5">
               Add your open-source or personal projects to showcase on your profile.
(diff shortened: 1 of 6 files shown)


dashboard/src/pages/MyProfile.vue
idlip, line 1: Also for two toggle functions:

https://github.com/fossunited/fossunited/blob/8f924d941b328829a1c5c213e6fb192a123dcb53/dashboard/src/pages/MyProfile.vue#L295-L333

can you please use DocumentResource method to toggle and save, so we can remove those two api functions.

https://github.com/fossunited/fossunited/blob/8f924d941b328829a1c5c213e6fb192a123dcb53/dashboard/src/pages/EventRsvpEdit.vue#L284-L293

idlip, line 1: Also use https://github.com/fossunited/fossunited/blob/8f924d941b328829a1c5c213e6fb192a123dcb53/dashboard/src/helpers/utils.js#L38-L45

for same func used in this page as well

idlip, line 1: https://github.com/fossunited/fossunited/blob/8f924d941b328829a1c5c213e6fb192a123dcb53/dashboard/src/pages/MyProfile.vue#L443-L457

Instead of this api call, can you make it so we directly use profileDoc to save those data in vue itself?

So we remove api function again.

For inspiration you can see: https://github.com/fossunited/fossunited/blob/develop/dashboard/src/pages/EventRsvpEdit.vue

I just think its clean.

In that case, instead of reactive profile_dict we simply v-model bind all form to its respective field (reduces code?)

dashboard/src/pages/MyProfile.vue
```diff
@@ -199,6 +199,132 @@
...
+                    type="url"
+                    placeholder="https://github.com/..."
+                  />
+                  <FormControl
+                    v-model="editingProject.tagline"
+                    label="Tagline *"
+                    placeholder="A short description"
+                    class="md:col-span-2"
+                  />
+                </div>
```
idlip, line 281: You forgot about `cover_image` field (optional, but yea)
https://github.com/fossunited/fossunited/blob/8f924d941b328829a1c5c213e6fb192a123dcb53/fossunited/foss_profiles/doctype/foss_user_projects/foss_user_projects.json#L24-L28

Also instead of binding each part to new or editing, do you think it'd be easy to have it in list view?
Like hackathon issue/PR is shown? 
https://github.com/fossunited/fossunited/blob/develop/dashboard/src/components/hackathon/HackathonProjectIssue.vue

I'm okay if its better to have unique design to present them (just wondering for simpler path for easy maintenance)

dashboard/src/pages/MyProfile.vue
```diff
@@ -293,85 +415,42 @@ const setProfileImage = (_file) => {
...
-        toast.info('Profile is now private')
-      } else {
-        toast.info('Profile is now public')
-      }
-      profile.fetch()
-    },
+  profileDoc.value.setValue.submit({ is_private: !profileDoc.value.doc.is_private }).then(() => {
+    toast.info(profileDoc.value.doc.is_private ? 'Profile is now private' : 'Profile is now public')
   })
 }
```
idlip, line 426: This does not work for me.

I think both v-model and this is clashing.
Instead of click on switch, add `@update`

this can be simple to

```js
# on swtich
@update:model-value="toggleProfilePrivacy"


const toggleProfilePrivacy = (value) => {
  profileDoc.value.setValue
    .submit({
      is_private: value,
    })
    .then(() => {
      toast.info(value ? 'Profile is now private' : 'Profile is now public')
    })
}
```

same can be applied on `toggleShowActivity`

fossunited/foss_profiles/doctype/foss_user_profile/templates/foss_user_profile.html
```diff
@@ -81,38 +81,22 @@ <h5>{{ section | capitalize }}</h5>
...
+            href="{{ project.project_link }}"
+            target="_blank"
+            rel="nofollow noopener"
+            data-docname="{{ project.name }}"
+          >
+            <div class="project-card-icon-wrap">
+              <i class="ti ti-folder-code"></i>
             </div>
-          </div>
+            <div class="project-title">{{ project.project_name }}</div>
```
idlip, line 98: final touch, show cover_image if exists as side icon?

idlip (changes requested):

idlip (approved):

idlip:
Well well, its works as expected and LGTM.

Thank you for applying changes as requested.

Feel good to reduce api endpoints and have some cruft removed.
Profile page also works well.

Just final change, the save button is odd toward left side.

image

Can we show that at bottom last?

image


Also the add project button can have better UX, its ideally shown right side or near projects heading so its easier to understand?
That icon looks left out and lonely :(

idlip:
Nice.

Thank you good to go!

Issue #1461, opened 2026-03-06
Improve UX for registering for a chapter's upcoming events if no events are active


agriyakhetarpal:
Hi! I'd like to "subscribe" myself to the activities of a chapter so that I can stay updated on them – i.e., I'd like to see the "Yes, I'd like to receive email updates about future events from <...chapter name here...>" check box somewhere in the UI.

I thought of opening this as an issue rather than a discussion because IMO the UX for this is not optimal – for example, for the [FOSS United March Community Call](https://fossunited.org/c/foss-united-organisation/community-call-march) that I ended up missing, I'd like to see one (or both) of the following options if I navigate to it today:
- The https://fossunited.org/c/foss-united-organisation/community-call-march page should mention something on the lines of "This event has concluded. Want to receive email updates about future events from <...chapter name here...>?"
- or https://fossunited.org/c/foss-united-organisation (the main chapter dashboard) should have a "Subscribe" option (primarily email, and perhaps RSS too via #1357) so that it is not necessary for one event to be active.

I'll still say it's better if you take these suggestions with a grain of salt, as I am by no means a UX expert :) It'd be nice if you could decide on the best approach forward among the team.

Thank you!

agriyakhetarpal:
Ideally, FWIW, the best UX would be to have some sort of "Subscriptions" heading in the sidebar panel under a user's profile at https://fossunited.org/dashboard/m,  so that a user can check what chapters they're subscribed to, and easily unsubscribe from/resubscribe to them with a single click (please consider this as a problem statement from my end for a FOSS Hack 2026 project!).

idlip:
Hi Agriya,

thanks for your feedback.

This makes sense to extend and add on top.

This was added based on request on #1012 so people can stay in touch with chapter.

But the current scenario I notice is, nobody is making use of this feature.
I can only see 1 email sent as chapter level mail. Now it becomes a false impression that the organizer will inform on future events.

As you said RSS would be better on each chapter too.
We already have rss for all upcoming events: https://fossunited.org/events/timeline/rss.xml

If this feature (chapter mailing/subscription) was used well, then we could have extended to show what all user has subscribed to (email group) and maybe show recent updates from them in dashboard (say timeline). Would have been cool.

For now yea, @ni5arga would you like to add rss for chapter's upcoming events?

ni5arga:
yes sure @idlip, I'll be glad to work on it.

agriyakhetarpal:
[...]

PR #1462, opened 2026-03-06
feat(rss): add rss feed for upcoming published events

ni5arga:
## Status

- [x] Ready for review

## Related Issue

Closes / Addresses: #1461 

## PR Description

Adds a per-chapter RSS feed at /c/{slug}/rss.xml listing upcoming published events.

- Feed includes event type, location, start/end time, and description; sorted by start date
- Autodiscovery added to each chapter page
- Extends shared templates/rss.xml base template

### DIFF
--- a/fossunited/chapters/doctype/foss_chapter/templates/foss_chapter.html
+++ b/fossunited/chapters/doctype/foss_chapter/templates/foss_chapter.html
@@ -85,7 +85,28 @@
     'fossunited/templates/macros/event_card.html' import event_card
   %}
   <div class="container my-5">
-    <h3 class="subtitle">Upcoming Events</h3>
+    <div class="d-flex align-items-center justify-content-between flex-wrap gap-2 mb-3">
+      <h3 class="subtitle mb-0">Upcoming Events</h3>
+      <div class="d-flex gap-2">
+        <a
+          href="/c/{{ doc.slug }}/rss.xml"
+          class="btn btn-sm btn-secondary"
+          target="_blank"
+          rel="noopener noreferrer"
+          aria-label="RSS feed for upcoming events"
+        >
+          <i class="ti ti-rss" aria-hidden="true"></i> RSS Feed
+        </a>
+        <button
+          id="download-ics-btn"
+          class="btn btn-sm btn-secondary"
+          data-event-ids="{{ upcoming_events | map(attribute='name') | list | tojson | e }}"
+          aria-label="Download upcoming events as ICS calendar file"
+        >
+          <i class="ti ti-calendar-down" aria-hidden="true"></i> Download ICS
+        </button>
+      </div>
+    </div>
     <div class="events-grid-4">
       {% for event in upcoming_events %}{{ event_card(event) }}{% endfor %}
     </div>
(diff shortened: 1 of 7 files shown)


fossunited/www/c/rss.py
```diff
@@ -0,0 +1,116 @@
...
+    host = get_request_site_address()
+
+    events = frappe.get_all(
+        EVENT,
+        filters={
+            "chapter": chapter.name,
+            "event_end_date": (">=", frappe.utils.now()),
+            "status": "Live",
+            "is_published": 1,
+        },
```
idlip: Please use constant for (CHAPTER)

Also if we get doc, it also has method func to get upcoming events.

```
doc = frappe.get_doc("FOSS Chapter", {"slug": "letsmakepy"})
doc.get_upcoming_events()
```

Since we use same formatting as upcoming events rss from timeline, can we reuse that as well?
Its less line of code only, but its repeating the same so.

fossunited/chapters/doctype/foss_chapter/templates/foss_chapter.html
idlip, line 1: LGTM.

Finally can you please add "Rss feed" button in the chapter page?

Also we've ics generator for events.

https://github.com/fossunited/fossunited/blob/8f924d941b328829a1c5c213e6fb192a123dcb53/fossunited/api/chapter.py#L97-L107

Can you also add another button for download ics ?
Simply pass ids from upcoming events for download.

idlip (changes requested):

fossunited/hooks.py
```diff
@@ -85,6 +85,7 @@
     {"source": r"c/(.+)/cfp", "target": r"/dashboard/cfp/apply/\1"},
     {"source": r"c/(.+)/cfp/all", "target": r"/dashboard/cfp/all/\1"},
     {"source": r"c/(.+)/schedule", "target": r"/dashboard/schedule/\1"},
+    {"source": r"c/([^/]+)/rss\.xml", "target": r"/c/rss.xml?chapter=\1"},
```
idlip: instead of regexp we can add this to `website_route_rules` above

```
    {"from_route": "/c/<chapter>/rss.xml", "to_route": "/c/rss.xml"},
```

fossunited/chapters/doctype/foss_chapter/templates/foss_chapter.html
```diff
@@ -113,10 +146,39 @@ <h3 class="subtitle">Organisers</h3>
...
+              a.click()
+              document.body.removeChild(a)
+              URL.revokeObjectURL(url)
+            }
+          },
+          always: function () {
+            btn.prop('disabled', false).html('<i class="ti ti-calendar-down" aria-hidden="true"></i> Download ICS')
+          },
+        })
+      })
```
idlip: Okay.. since we are using this download button third time here.
Can we simply make macro on it so in 3 pages we can simply use via passing ids?

idlip (changes requested):

idlip:
Thank you for making the changes.

Works as expected.

Maybe in future we can direct file download response from the api

```

    ics_data = c.serialize()

    frappe.response.filename = "events.ics"
    frappe.response.filecontent = ics_data
    frappe.response.type = "download"
    
```


and simply call api directly via href, in case we want to avoid js script running on this.

PR #1467, opened 2026-03-09
fix(dashboard): sanitize HTML in cleanedHTML helper using DOMPurify

### LINKED ISSUE #1165: Dom purify & Security Check
- We need some security analysis on platform

- https://github.com/fossunited/fossunited/pull/1164#discussion_r2360539199

- Optimize code in all page

Anshi321:
This PR adds HTML sanitization to prevent potential XSS issues when rendering dynamic content.

The  cleanedHTML() helper in utils.js now uses DOMPurify.sanitize() before parsing HTML with  tempDiv.innerHTML . This ensures that any unsafe tags or scripts are removed before the content is processed.

## Changes Made
- Added DOMPurify.sanitize() in the cleanedHTML() helper
- Ensured sanitized HTML is used when rendering `review.remarks`
- Prevented raw unsanitized HTML from being rendered via v-html in ReviewSection.vue

## Why this change?
Previously, HTML content was inserted using `tempDiv.innerHTML` without sanitization, which could potentially allow unsafe HTML to be rendered. Using DOMPurify ensures the content is cleaned before rendering.

## Issue
Closes #1165

### DIFF
--- a/dashboard/src/components/reviewers/ReviewSection.vue
+++ b/dashboard/src/components/reviewers/ReviewSection.vue
@@ -39,7 +39,7 @@
             <div
               v-if="review.remarks"
               class="prose prose-sm max-w-full"
-              v-html="review.remarks"
+              v-html="cleanedHTML(review.remarks)"
             ></div>
             <span v-else class="text-sm text-ink-gray-5">No Remarks</span>
             <div class="flex items-center gap-2">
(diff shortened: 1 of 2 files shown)


idlip:
Ideally speaking, remarks can be only added by "CFP Reviewers" who has permission.

We've around 40 reviewers who are known and trusted in community.
But nonetheless, PR adds sanitization directly to v-html content, so yeah a good start to keep pages clean.

Thank you for adding.

Are you going to add more commits in this PR to cover more pages or will it be later on another PR?

Anshi321:
Thanks for the feedback!

For this PR I wanted to keep the change small and focused on securing the cleanedHTML() helper first. My plan is to address the other v-html / HTML rendering locations in separate follow-up PRs so each change stays easier to review.

I’ll open additional PRs to cover those pages as well.

Issue #730, opened 2024-12-18
[FEATURE] [metaissue] Improve performance


rahulporuri:
### Feature description

We measure a bad (51/100) on mobile for performance on google pagespeed insights - https://pagespeed.web.dev/analysis/https-fossunited-org/997lz0wo6r?form_factor=mobile. it will be good to look into what we can do to improve performance of the website in this regard. Desktop performance is relatively better at 88/100 but it could still be improved.

rahulporuri:
another useful tool to measure/understand performance is https://www.webpagetest.org - see the results of this test https://www.webpagetest.org/result/241218_BiDcXR_7WH/. https://developer.mozilla.org/en-US/docs/Learn/Performance/Measuring_performance was quite useful to understand where we could potentially start, given that we ourselves don't have prior experirince with improving performance

rahulporuri:
FTR @pecpo, @xegtor, and @swapnil-panigrahi will be working on this project overall.

ni5arga:
hi! i would like to work on this issue - the score still seems to be around 51-52 for mobile devices.

https://pagespeed.web.dev/analysis/https-fossunited-org/o0yryri6h3?form_factor=mobile

we can start by optimizing the custom.css file according to the recommendations.

Issue #1421, opened 2026-02-12
[Meta] Reduce JS and css dependency


idlip:
Frappe already bundles huge css and JS files.

Our custom.css is expanding to 4K Lines, although many lines can be removed.
frappe already has bootstrap v4 css file embedded and many others which we can use more programmatically.
Now its a battle of do we need control over our defined rules (vanilla which serves our purpose) or since the other files are imported, do we just utilize them as its offered.

Imo as code is growing, we must reduce re-creating wheel and next step should be taken to yak-shave css rules here and our css must only serve for color scheme implementation (dark mode and more)
- Jeswin had idea of adding accent colors for each event (as organizers choose) so glow up (like candy?) for pages

- This can be big changes, but rewarding for future maintenance.

Same with JS files, the bundle file (website.bundle.js) imports so many frappe scripts.
I think we are not using much of these codes, but yet imported ?
Again, this can be made useful if anywhere its applicable. Most of our pages are simply static, but recently [grants](https://fossunited.org/grants/) suite of pages has JS for searching and filtering (so that's a place to look for using if frappe JS can suffice)

- Lazy load if possible or only load wherever its required.

----

This is to review and scan the codebase so we can improve performance bit by bit.

aflaha01:
Hi, I’m new here and would love to work on this issue. Please let me know if I can take this up

idlip:
> Hi, I’m new here and would love to work on this issue. Please let me know if I can take this up

Hi, please feel free to do this in stages. Like page by page for each PR.
Frappe bundles so many js files and css (bootstrap v4). Other plan was to use tailwind itself #692, but since we already import these bulk files from frappe side, its better to use them as much as possible and try to only write utility based css for custom use.

reference:
- https://colton.dev/blog/tailwind-is-the-worst-of-all-worlds/
- https://www.zolkos.com/2025/12/03/vanilla-css-is-all-you-need
- https://dev.37signals.com/modern-css-patterns-and-techniques-in-campfire/
- https://moderncss.dev/

aflaha01:
Thanks for the clarification!
Just to confirm  should I prioritize refactoring pages to better utilize existing Bootstrap (from Frappe) and reduce custom CSS, or should I move in the direction of rewriting templates in Tailwind as mentioned in #692?
I can start with one simple page and raise a PR for review before proceeding further

idlip:
[...]

PR #1468, opened 2026-03-09
feat(performance): load fonts non-blocking and remove legacy font-face formats

ni5arga:
- Replaced 3 render-blocking `@import` font calls in custom.css in with a single non-blocking `<link rel="preload">`  in foss_base.html

- Merged duplicate Inter imports into one consolidated Google Fonts URL

- Dropped legacy @font-face formats (.eot, .ttf, .svg) from FFF Forward; kept woff2 + woff only

- Added font-display: swap to `@font-face`

- Removed unused --ff-serif CSS variable (Source Serif Pro was never loaded)

Render-blocking `@import` statements inside CSS files force the browser to pause page rendering until all font stylesheets are downloaded. This was causing an estimated ~1,200ms render-blocking delay on every page load. Additionally, the missing `font-display: swap `was causing an extra ~140ms penalty.

Related issue: #730

### DIFF
--- a/fossunited/public/css/custom.css
+++ b/fossunited/public/css/custom.css
@@ -13,8 +13,6 @@
 :root {
   /* font-families */
   --ff-san-serif: 'Inter', sans-serif;
-  --ff-mono: 'Source Code Pro', monospace;
-  --ff-pixel: 'Press Start 2P', cursive;
 
   /* font-sizes */
   --text-xs: 0.75rem;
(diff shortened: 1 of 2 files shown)


idlip:
Actually frappe bundles Inter font in `frappe/public/css/fonts/inter/InterVariable.woff2`

http://foss.localhost/assets/frappe/css/fonts/inter/Inter-Thin.woff2

Can we use that directly.

We can remove source code pro, i cannot find anywhere it being used.

ni5arga:
`--ff-pixel (Press Start 2P)` isn't being used anywhere as well, removing it too.

idlip:
Shouldn't we define `@font-face` to say where is inter?

ni5arga:
If I am not wrong, the `@font-face` for Inter is already declared globally by frappe before the template even renders.

idlip:
oh yes it bundles loads of things.

#1421

PR #1471, opened 2026-03-09
feat(profile): add image size limit and handle decompression bomb errors

ni5arga:
- Handle decompression bomb errors  

- Reject uploads larger than 5 MB before any processing occurs

- Add per-user rate limit (10 reqs/5 minutes) to set_profile_image and set_cover_image

### DIFF
--- a/fossunited/api/profile.py
+++ b/fossunited/api/profile.py
@@ -1,12 +1,15 @@
 import io
 
 import frappe
+from frappe.rate_limiter import rate_limit
 from frappe.utils.file_manager import save_file
 from PIL import Image
 
 from fossunited.api.dashboard import get_session_user_profile
 from fossunited.doctype_ids import CHAPTER, RESTRICTED_USERNAME, USER_PROFILE
 from fossunited.fossunited.utils import sanitize_text_content
+
+MAX_IMAGE_SIZE_BYTES = 10 * 1024 * 1024
 
 
 def convert_image_to_webp(image_content: bytes) -> bytes:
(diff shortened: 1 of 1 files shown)


fossunited/api/profile.py
```diff
@@ -19,11 +23,14 @@ def convert_image_to_webp(image_content: bytes) -> bytes:
...
             img.save(webp_io, format="WEBP", quality=80)
             return webp_io.getvalue()
+    except Image.DecompressionBombError:
+        frappe.throw(_("Image exceeds maximum allowed dimensions."))
     except Exception as e:
         frappe.throw(f"Failed to process image: {str(e)}")
 
 
 @frappe.whitelist()
+@rate_limit(key="user", limit=10, seconds=60)
```
idlip: I think 1 per 10 secs is still small?

Shouldn't this be more stricter?

for 5 or 10 mins?
Imo can go to an hour as well. Since why would someone keep changing photo >5 times.

fossunited/api/profile.py
```diff
@@ -1,11 +1,15 @@
...
 import frappe
+from frappe.rate_limiter import rate_limit
 from frappe.utils.file_manager import save_file
 from PIL import Image
 
 from fossunited.api.dashboard import get_session_user_profile
 from fossunited.doctype_ids import CHAPTER, RESTRICTED_USERNAME, USER_PROFILE
+from fossunited.fossunited.utils import sanitize_text_content
+
+MAX_IMAGE_SIZE_BYTES = 10 * 1024 * 1024
```
idlip: Lets keep it 5 MB.

fossunited/api/profile.py
ni5arga: Changing it to 10 reqs per 5 minutes.

idlip:
So quick!

Thank you!

PR #1474, opened 2026-03-09
fix: pin tatsu to <5.10.0 to resolve install crash

jasilfaras:
## Status

- [ ] Draft
- [x] Ready for review

---

## Related Issue

Closes / Addresses: N/A

---

## Problem

New installations of the `fossunited` app were failing during the DocType synchronization phase. The crash occurred because the `tatsu` library recently released version `5.10.0+`, which removed legacy Python compatibility layers (specifically `generic_main`). This caused a `ModuleNotFoundError` that blocked the entire `bench install-app` process.

---

## Solution

Pinned the `tatsu` version to `<5.10.0` in `pyproject.toml` and updated the `uv.lock` file. This ensures that the environment installs a safe version of Tatsu (like `v5.9.2`) that still contains the necessary modules for Frappe to sync DocTypes correctly.

---

- [x] Tested locally
---

## Changelog

fix: pin tatsu to <5.10.0 to resolve install crash

---

## Visualization

image

---

## Further Ahead

This is a temporary stability fix. The codebase or underlying Frappe framework may eventually need to be updated to be compatible with Tatsu `5.10+`, but this unblocks local development and installations immediately.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,7 @@ dependencies = [
     "Pillow~=11.0.0",
     "ics~=0.7.2",
     "pre-commit>=4.2.0",
+    "tatsu<5.10.0",
 ]
 
 [project.optional-dependencies]
--- a/uv.lock
+++ b/uv.lock
@@ -32,15 +32,6 @@ wheels = [
 ]
 
 [[package]]
-name = "babel"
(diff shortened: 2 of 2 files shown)


idlip:
Thank you

PR #1476, opened 2026-03-10
feat(events): show message when event has concluded

Anshi321:
Adds a small UX improvement to event pages.

If an event has already concluded, a message "This event has concluded." is displayed below the event title. This helps users quickly understand that the event is no longer active when visiting the page.

The change is minimal and only introduces a conditional check in the event template to show the message when the event status is concluded.

Related to #1461

### DIFF
--- a/fossunited/chapters/doctype/foss_chapter_event/templates/foss_chapter_event.html
+++ b/fossunited/chapters/doctype/foss_chapter_event/templates/foss_chapter_event.html
@@ -112,6 +112,12 @@
       Registration is closed for the event
     </div>
     {% endif %}
+    {% if status_concluded %}
+    <div class="v3-status v3-status--warning" role="status" aria-live="polite">
+      <i class="ti ti-ban" aria-hidden="true"></i>
+      This event has concluded.
+    </div>
+    {% endif %}
 
 
   </div>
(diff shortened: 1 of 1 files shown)


Anshi321:
The CI failure seems unrelated to this change since this PR only modifies the event page HTML template by adding a conditional message when an event has concluded. No backend logic was changed.

Could you please take a look? Happy to make any adjustments if needed.

idlip:
> The CI failure seems unrelated to this change since this PR only modifies the event page HTML template by adding a conditional message when an event has concluded. No backend logic was changed.

Dont worry, sometimes that test fails randomly on CI. Idk why.

Currently the text looks in <p> in between:

image


I'd be better to show up in status tag of event_cover macro?

https://github.com/fossunited/fossunited/blob/a5c712a5a5b0a72a0c402d61d874e01dbf890ace/fossunited/chapters/doctype/foss_chapter_event/templates/foss_chapter_event.html#L109-L114

Just like we've registration tags

image

Alongside those.

Anshi321:
Thanks for the suggestion! I’ll update the implementation to show the concluded message as a status tag alongside the other event status tags.

idlip:
@Anshi321 

Can you remove the trailing fullstop (.) ?
Looks odd

image

Gtg, or i can modify it later as well

Anshi321:
Got it, thanks! I'll remove the trailing full stop and push the update.

idlip:
Thank you!

Anshi321:
Thanks for the review and suggestions!  
Glad to see the change merged. Happy to contribute.

Issue #1364, opened 2026-01-08
Delete an event using dashboard


fossdot:
As an organizer/city chapter manager, I can't delete an event an event.

idlip:
Is this really something that we need?

I mean, for anyone who made event open and if some rsvp or cfp were filled then it might cause issue of missing event doctype link.

agriyakhetarpal:
Hi!

As a virtual chapter manager/organiser, I, too, would like this. If the event were cancelled by an organiser or they would want it to be removed from the chapter page for any reason (perhaps copyright or it has information that should not be public), then it would be nice if the platform could allow deleting the event from the list, and emails would be sent out with some templated message "This event has been cancelled. You may remove its details from your calendar" and any additional information.

idlip:
for now, I've added unpublish event to toggle publishing of event page.

We can add delete event with condition tho, like if event is very new (created within 6 days) this option can be shown..?

Anshi321:
Hi, I’d like to work on this.

Based on the discussion, it seems better to keep deletion limited rather than allowing it in all cases.

I’m planning to:
- show the delete option only if the event is recently created (around 5–6 days)
- block deletion if there are existing RSVPs/CFPs linked to the event

Let me know if this approach works or or would you prefer any changes to this logic.

idlip:
Hi @Anshi321 

That sounds about right.
Please go ahead and try minimal implementation using frappe.client.delete itself.

Since we've `const event = createDocumentResource` event data in the EventDetails.vue page, you can use that and add button.
- But for rsvp or cfp, we might need to gather it someway

Still I think it'd be better for us (system manager) to delete in desk directly.

Edit: Accidentally closed it lol

Anshi321:
Thanks! I’ll proceed with a minimal implementation using frappe.client.delete and add the button in EventDetails.vue.

agriyakhetarpal:
Coming back to this and reflecting on my thoughts, I now think it would be better to "Archive" an event instead and show it in a separate "Archived" events list, only visible to the chapter organisers. Fully deleting an event isn't useful because it can remove useful history for later; meanwhile, putting it in another basket that I don't need to access often could be a good idea. GitHub itself is experimenting with ways to archive PRs in a repository so that they are only visible to the repository's maintainers or the org administrators.

[...]

PR #1491, opened 2026-03-19
docs: clarify event deletion and archival behavior

Anshi321:
Adds a clarification under the event management section about how event deletion works.

Events cannot be deleted or archived directly from the platform. The docs now mention using "Unpublished", "Draft", or "Cancelled" to hide events, and reaching out to developers@fossunited.org if permanent removal is required.

Related to #1364


idlip:
Thank you for adding them.
LGTM!

Anshi321:
Thanks a lot! Glad this helps. Appreciate the review.

PR #1497, opened 2026-03-22
refactor: replace custom CSS with Bootstrap and clean JS in grantees …

aflaha01:
## Status

- [ ] Draft
- [x] Ready for review

## Related Issue

Addresses: #1421

### Summary
Refactored the grantees page to reduce dependency on custom CSS and better utilize Bootstrap utilities already available via Frappe.

### Changes
- Replaced `v3-btn` with Bootstrap `btn` classes
- Removed usage of `hidden` class and switched to `d-none`
- Reduced usage of `v3-*` classes in JS-rendered sections
- Minor JS cleanup (null checks, safer DOM handling)

### Notes
No functional changes are intended.  
There may be slight visual differences due to replacing custom styles with Bootstrap equivalents.

I was not able to fully verify all UI states across devices due to lack of complete local Frappe setup, so feedback is welcome and I’m happy to refine based on review.

## Further Ahead

- Explore replacing more custom styles with Bootstrap utilities
- Review JS for potential performance improvements or reuse of Frappe utilities

### DIFF
--- a/fossunited/www/grants/grantees/index.html
+++ b/fossunited/www/grants/grantees/index.html
@@ -8,16 +8,16 @@
 {% endblock %}
 
 {% block action_buttons %}
-  <a href="/grants/projects" class="v3-btn v3-btn-secondary">
+  <a href="/grants/projects" class="btn btn-outline-secondary me-2 mb-2">
     <i class="ti ti-device-imac-code"></i> Project Grants
   </a>
-  <a href="/grants/events" class="v3-btn v3-btn-secondary">
+  <a href="/grants/events" class="btn btn-outline-secondary me-2 mb-2">
     <i class="ti ti-building-circus"></i> Event Grants
   </a>
-  <a href="/grants/fellowships" class="v3-btn v3-btn-secondary">
+  <a href="/grants/fellowships" class="btn btn-outline-secondary me-2 mb-2">
     <i class="ti ti-heart-handshake"></i> Fellowships
   </a>
-  <a href="/grants/grantees/rss.xml" class="v3-btn v3-btn-secondary" title="RSS Feed">
+  <a href="/grants/grantees/rss.xml" class="btn btn-outline-secondary me-2 mb-2" title="RSS Feed">
     <i class="ti ti-rss"></i> RSS
   </a>
 {% endblock %}
--- a/fossunited/www/grants/grantees/index.html
+++ b/fossunited/www/grants/grantees/index.html
@@ -8,16 +8,16 @@
 {% endblock %}
 
 {% block action_buttons %}
-  <a href="/grants/projects" class="btn btn-outline-secondary me-2 mb-2">
+  <a href="/grants/projects" class="btn btn-outline-secondary mr-2 mb-2">
     <i class="ti ti-device-imac-code"></i> Project Grants
   </a>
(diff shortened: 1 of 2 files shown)


idlip:
We miss button theme and card gaps.

- before (default)
image


- After (this PR)
image


Is there a way we can inform bootstrap to use variables colors for its naming schema? (btn secondary and so on?

- Also toggle section does not work. maybe class mismatch?

- https://github.com/fossunited/fossunited/blob/68f0907443d5e0f199fc8c1cdb5f78803b0a5af6/fossunited/public/js/common_functions.js#L318-L325

You can use this common function for toggle and remove toggleYear in this file

aflaha01:
Hi @idlip Can you take a look when you’re free and let me know if anything needs to be changed? Thanks :)

fossunited/www/grants/grantees/index.html
idlip, line 1: No need any changes (margin) in this page.

fossunited/www/grants/grantees/index.js
```diff
@@ -1,71 +1,67 @@
 let originalItems = []
 
 document.addEventListener('DOMContentLoaded', () => {
```
idlip, line 3: Can I get TLDR why not clone?
Since we do grouping and sorting, wouldn't it be better?

fossunited/www/grants/grantees/index.js
```diff
@@ -123,49 +125,63 @@ function applyFilters() {
...
+    <div class="d-flex flex-column" id="${sectionId}"></div>
+  `
+
+  const content = section.querySelector(`#${sectionId}`)
+  items.forEach((item) => {
+    item.classList.add('mb-3')
+    content.appendChild(item)
+  })
+
   document.getElementById('grants-container').appendChild(section)
```
idlip, line 182: Can we retain this back to prev one.

As you see in screenshot i attached above, there is visual changes.

In dark mode text is muted to gray.
For toggle to work we need to remove "d-flex flex-column" from 170th line

Since you remove section, in dark mode text is not visible. And you've added border-bottom, but default has bottom center aligned along with title (heading) as:

Heading ------------------- v

Ik you don't setup to test locally.
But can you please only inherit JS improvements that you are doing without affecting css class/rules (that can be diff PR, where we 1-1 replace css rules with bootstrap rules)

idlip (changes requested):

idlip:
hey @aflaha01 there is some visual changes with css rules there.

I forgot to add previous site screenshot: https://fossunited.org/grants/grantees

image

^ This is how default looks. So can you revert back all css changes and only touch js level changes now?

Css changes needs to be 1-1 replace so that can be diff PR on its own.

aflaha01:
Hi @idlip can u take a look

idlip:
Hi @aflaha01 

Did you take a look at all my review comments?

I dont think you took care of everything.
Can you please go through it once and close them as you fix or make that changes.

1. No need any change on fossunited/www/grants/grantees/index.html file

2. There is still visual changes. You only reverted in grouping by year, others have changes left.

IMO it'd be better to start off from scratch changing in small bits.

idlip:
Closing this due to conflicting files. Feel free to re-open by rebasing

PR #1509, opened 2026-03-25
feat(cfp): add "Edit proposal" button for proposal owner on public page

### LINKED ISSUE #1502: The public page for a proposal should display an "Edit proposal" button to an authenticated user
Hi!

With the current flow of the platform, a user on the public page for a proposal they submitted and who wants to edit their submission cannot do so directly from that page. They need to navigate to the dashboard from the profile picture dropdown, select "My Proposals", choose the proposal specifically to edit, and then make the necessary changes. There are a lot of clicks involved with this process.

It would probably be better if:
- (short/quick fix) –> if the proposal submitter matches the current authenticated user (i.e., it is the user's proposal), there should be an "Edit proposal" option that points to the proposal directly in the dashboard and the browser history should reflect this such that the back button points to the public proposal page instead of moving to the dashboard
- (longer, happy to open this as another issue if you are interested) -> it would be better to edit the proposal in-place; i.e., that the behaviour should be similar to how the description of a GitHub issue (like this one) can be edited, instead of having to navigate to an entirely different page altogether.

Anshi321:
## Summary

On the public proposal page, the proposal owner didn’t have a direct way to edit their submission and had to navigate through the dashboard.

## Changes

- Added a check in `get_context` to verify if the logged-in user matches `submitted_by`, `email`, or speaker email
- Passed `is_owner` and `proposal_id` to the template context
- Conditionally rendered an "Edit proposal" button for the proposal owner
- Redirects to `/dashboard/my-proposals/edit/<proposal_id>`

## Closes

Closes #1502

### DIFF
--- a/fossunited/fossunited/doctype/foss_event_cfp_submission/templates/foss_event_cfp_submission.html
+++ b/fossunited/fossunited/doctype/foss_event_cfp_submission/templates/foss_event_cfp_submission.html
@@ -50,7 +50,7 @@
       <h1 class="text-3xl font-bold">{{ doc.talk_title }}</h1>
       <div class="flex gap-2 items-center">
         {% if is_owner %}
-          <a href="/dashboard/my-proposals/edit/{{ proposal_id }}" class="btn btn-default btn-sm">Edit proposal</a>
+<a href="/dashboard/my-proposals/edit/{{ doc.name }}" class="btn btn-default btn-sm">Edit proposal</a>
         {% endif %}
         {{ like_button() }}
       </div>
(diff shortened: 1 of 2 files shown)


fossunited/fossunited/doctype/foss_event_cfp_submission/templates/foss_event_cfp_submission.html
```diff
@@ -48,7 +48,12 @@
     </div>
     <div class="flex justify-between gap-4">
       <h1 class="text-3xl font-bold">{{ doc.talk_title }}</h1>
-      {{ like_button() }}
+      <div class="flex gap-2 items-center">
+        {% if is_owner %}
+          <a href="/dashboard/my-proposals/edit/{{ proposal_id }}" class="btn btn-default btn-sm">Edit proposal</a>
```
idlip: You can directly refer proposal_id as doc.name

Doc is whole doc fields from doctype

Anshi321: Thanks! updated it to use doc.name directly and removed the extra context.

idlip:
Thank you, LGTM.

Anshi321:
Thanks for reviewing and merging!

PR #1524, opened 2026-03-26
fix: align Activity section and restore top border

### LINKED ISSUE #1520: Improper styling for the Activity section on public profile pages
Hi!

Image

Looks like this section on users' public profiles could do with some love. Particularly:
- "Activity" and the sentence below it are misaligned
- There is a slight gray border on the bottom, left, and right sides, but it doesn't extend on the top border

Anshi321:
Closes #1520

Fixed the alignment issue in the Activity section and restored the missing top border.

- Removed `border-top-0` so the top border is visible again
- Adjusted heading padding to align it with the content below

Kept the changes minimal and consistent with the existing layout.

Before:
Screenshot 2026-03-27 111613
 
After:
Screenshot 2026-03-27 112114

### DIFF
--- a/fossunited/foss_profiles/doctype/foss_user_profile/templates/foss_user_profile.html
+++ b/fossunited/foss_profiles/doctype/foss_user_profile/templates/foss_user_profile.html
@@ -48,7 +48,7 @@
 
       <div class="profile-content-section mt-10">
         <div class="d-flex align-items-center">
-          <h5 style="padding: 2rem 2.5rem;">
+          <h5 style="padding-left: 2.5rem;">
             Activity
           </h5>
           <a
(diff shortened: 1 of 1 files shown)


idlip:
Hi,

Actually the "user has choose..." can be shift right.

Since other info cards with some indentation (margin)

image

So yea, need to adjust with both to be aligned at same point.

- Also maybe border can be removed as well

Anshi321:
Thanks for the suggestion! I’ve updated the alignment to match the other sections and adjusted the border as discussed. Please have a look.

idlip:
image


Padding is set to 4 sides, and for "User has chosen..." that needs to be pushed to right to match the alignment.

With activity cards as well there is little misalignment.

image

idlip:
Thank you, should be fine.

PR #1526, opened 2026-03-27
fix: sync dark mode transition between sidebar and main area

### LINKED ISSUE #1516: Sidebar and main area on profile dashboard switch to dark/light mode out of sync with each other?
Hi!

The components on the profile dashboard at https://fossunited.org/dashboard/me, i.e., the sidebar and main area, switch between light and dark modes out of sync with each other. In particular, the sidebar appears to be slower to make the transition? It requires 0.3 seconds, and the other needs 0.15 seconds. Adjusting it to be 0.15 seconds still leaves some artifacting during the transition, though.

Compare this with something like https://fossunited.org/hack/fosshack26, and the act of switching to dark mode is more streamlined there.

I don't have a video right now, but happy to record one; thanks!

Anshi321:
Closes #1516

The sidebar had a transition applied while the main content switched instantly, causing a visible delay during theme toggle.

Added a matching transition to the main content so both switch consistently.

Kept the change minimal.

### DIFF
--- a/dashboard/src/pages/Home.vue
+++ b/dashboard/src/pages/Home.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="flex flex-col md:flex-row">
     <SideNavbar v-if="navItems.data" :menu-items="navItems.data" />
-    <div class="flex-1 min-w-0">
+    <div class="flex-1 min-w-0 transition-colors duration-300">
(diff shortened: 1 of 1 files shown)


idlip:
Oh shii... Thanks for noticing!

Issue #1546, opened 2026-04-09
Investigate Roles and Permissions required to provide reviewers access to proposals via desk


rahulporuri:
We have been under utilizing the Frappe Framework, specifically Roles and Permissions available in the Framework. Over the coming week, we should understand how Roles and Permissions work in the Frappe Framework, and come up with the necessary Roles and Permissions to enable Reviewers to access Proposals via the Frappe Desk (fossunited.org/app).

The expected user experience is as follows

- Reviewer has a FOSS United account
- Reviewer is given access to the desk (fossunited.org/app)
- Reviewer is only able to access proposals via desk, i.e., "FOSS Event CFP Submission" documents
- Reviewer has read-only access to the proposals via desk
- Reviewer is able to comment on the proposal via the inbuilt comment functionality on documents

PR #1553, opened 2026-04-16
feat: allow reviewers to access desk to make use of frappe's resources for organization

idlip:
- Frappe already has many bells and whistles (as itself calls it battery included)
- we miss on introducing much of these battery to platform users since they cannot use desk, fundamentally should've been solved with adding robust roles and permissions and as well sharing each doc with teams as they are created.

- Based on #1546 

- We will initially test with allowing especially Reviewers to access desk to add reviews for each cfp submission, since we can use "commenting" feature, assigning many reviewer for each proposal and so on in desk itself rather adding more code to dashboard.

- Since review table perm in elevated, have added js file so other fields are read-only and and only they can do reviews
- added controller so only one review can be added by them and only edit their own review (did not find any other easy way for child table :) )

- Note: Since many doctype have "All" permission, reviewers can still read those all in desk

### DIFF
--- a/fossunited/fossunited/doctype/cfp_submission_speaker/cfp_submission_speaker.json
+++ b/fossunited/fossunited/doctype/cfp_submission_speaker/cfp_submission_speaker.json
@@ -35,6 +35,7 @@
       "fieldtype": "Data",
       "in_list_view": 1,
       "label": "Email",
+      "permlevel": 4,
       "reqd": 1
     },
     {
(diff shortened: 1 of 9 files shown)


idlip:
Ok, test again (3rd time) and seems to be in right direction if we are launching desk access

Although reported a frappe bug (claude explained), thus in UI should be read-only for all base field, and reviewers can only add review and only once (can edit their own too).

Will merge with few linting fix.

Still on prod we've not enabled "desk access" for any role.
Theoretically desk access reflects same as "Permission" table for each doctype, if we've screwed those tables then desk also shows the same.

Will merge with few lint fixes

PR #1559, opened 2026-04-22
feat(doctype): add user approval to create foss_user_profile

idlip:
- We've been only getting spam account creation running ads via user profile page at `/u/<username>`
  All new accounts have been that

- Frappe does not have a way to do approval basis signups for "user"

- So we add simple controller to disable by default and add buttons to approve or deny user (system manager via desk)
- and they get email to await for approval

- workflow is:

  user -> disabled by default (with nothing to do)
  approve -> enabled + foss_user_profile is created
  denied -> user doctype is deleted (with optional email)

image

- "spam" button allows us to delete both user profile and user doctype 
  only used for spam accounts signed up
  
Note: No reinventing of wheel, just added convenience controller and button to access quick.

Added tests to make sure functionality remains same

### DIFF
--- a/fossunited/foss_profiles/doctype/foss_user_profile/test_foss_user_profile.py
+++ b/fossunited/foss_profiles/doctype/foss_user_profile/test_foss_user_profile.py
@@ -26,7 +26,7 @@ class TestFOSSUserProfile(FrappeTestCase):
 
     def test_private_profile_access(self):
         test_name = fake.name()
-        test_user = frappe.get_doc(
+        user_doc = frappe.get_doc(
             {
                 "doctype": "User",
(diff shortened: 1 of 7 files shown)


harshtandiya:
Wouldn't it be easier to enable 2 factor auths?

idlip:
> Wouldn't it be easier to enable 2 factor auths?

damn, beats me why i did not see that.

Will probably enable it then. Thanks for pointing out

idlip:
Sinking this for now.

We still get 20-30 new user sign ups, its tedious to enable and deny all.

Bigger plan is you only get profile picture when you are volunteer or someone participating in this community. If not no purpose in having profile as well.

Issue #1416, opened 2026-02-11
[Tests] Make more tests/utils.py to handle deleting of all created docs


idlip:
As things grow, unit tests are getting complicated as some method might not work as intended (edge case)

Scan all unittests to make sure we avoid building things there, lets handle creation and deletion via some utils function which will take care deterministically.

- So finally test files remain as simple as 

self.... change something and we expected this output.

```
doc.desc = "changed"
with self.assertRaises(frappe.PermissionError):
    doc.save()
```

harshtandiya:
I would suggest integrating https://github.com/TomasBarry/frappe_factory_bot for writing better / faster tests

I am trying to integrate this in Forms Pro too, once I do that, we can follow similar steps for this repo

idlip:
Nice. Thanks for bringing this up.

Just glanced at it, naming seems confusing. Gotta test what is "Factory boy" and test this to see if helps.

harshtandiya:
https://github.com/BuildWithHussain/forms_pro/pull/72 @idlip something you should look at too

idlip:
Yes, thanks for again pointing out.
Just yesterday I was looking at T4GC using cypress for whole frappe tests (bloated for our usecase, but it did frontend test as well)

Will keep an eye and implement it sooner!

PR #1560, opened 2026-04-22
feat(tests): add factories for FOSSChapter, FOSSChapterEvent, and RSVP

harshtandiya:
- Introduced factory classes to streamline the creation of test data for chapters, events, and RSVPs.
- Updated test cases to utilize new factories for improved readability and maintainability.
- Added a new CI step to install the frappe_factory_bot application.

## Status

- [ ] Draft
- [x] Ready for review

---

## Related Issue

Addresses: #1416 

---

## Problem

Streamlining tests for better readability, maintainability and ease of writing them
---

## Solution

Using a factory system to create documents on tests. A very bad version of it exists in codebase at tests/utils.py. This is much better. 

https://github.com/harshtandiya/frappe_factory_bot

---

- [x] Tested locally
 

---

## Changelog

<commit>: 

---

## Visualization

---

## Further Ahead

Refactoring existing tests and getting rid of methods in test/utils.py

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,6 +85,7 @@ jobs:
       - name: Install App and Setup Site
         working-directory: ${{ env.BENCH_PATH }}
         run: |
+          bench get-app harshtandiya/frappe_factory_bot --branch main
           bench get-app fossunited $GITHUB_WORKSPACE
           bench setup requirements --dev
           bench new-site --db-root-password root --admin-password admin ${{ env.SITE_NAME }}
--- a/fossunited/chapters/doctype/foss_chapter/test_foss_chapter.py
+++ b/fossunited/chapters/doctype/foss_chapter/test_foss_chapter.py
@@ -1,98 +1,69 @@
 import frappe
-from faker import Faker
 from frappe.tests.utils import FrappeTestCase
 
-from fossunited.doctype_ids import CHAPTER, USER_PROFILE
-from fossunited.tests.utils import insert_test_chapter
-
-fake = Faker()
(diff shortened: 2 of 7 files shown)


idlip:
hey @harshtandiya 

Thanks for initiating this.

Frappe factory adds minimal abstraction (helper) to have tests clean, which seems good to include.

https://github.com/search?q=frappe_factory_bot&type=code

From the quick code search, looks like only your project is making use of it? (fossunited will be second if so)

- Can you shed more insights, as in, will you be maintaining fork to improvise it? or is that base_factory.py file already is sufficient enough to directly copy it to project repo itself? (since thats the only file)

- But still the code feels alien to me initially (as of now).

Even I noticed erpnext which had rewrite for all tests, still use base method of creating and tearing down. 

I'm just not sure if its right call since others are not using it.

definitely looks helpful. Can you please write more docs on using it in your repo?

thanks

harshtandiya:
> Can you shed more insights, as in, will you be maintaining fork to improvise it? or is that base_factory.py file already is sufficient enough to directly copy it to project repo itself? (since thats the only file)

I plan on maintaining it for now. It's self sufficient as it is right now as well. We plan to use it increasingly at BWH too. I would suggesting using the repo as the app instead of copying the code, since any new updates i make will be lost if you copy the code. Anyways this app only needs to be installed in CI while running the tests, so better to use the app.

> But still the code feels alien to me initially
https://www.geeksforgeeks.org/python/factory-method-python-design-patterns/
A decent read

---

The idea in introducing this is to debloat the tests of this repo with factories and write more comprehensive tests (easily)

idlip:
Thanks for confirming.
If the OG developer says so 😜, time to clean code then!

PR #1578, opened 2026-05-06
test(bruno): add script to generate bruno collection for all api endpoints

idlip:
- simply read all whitelisted functions and generate yaml files in opencollection format

- after 3.1 they suggest OAS than bru file, so lets directly do that

- script also writes environment file if not exists, so adds all func params to fill for environment
  Note: We've same param repeating for many functions, so we need to stabilize or modify each yml
  to have its own param value

- in .env you've to fill auth token from frappe setting.

Note: I've not yet added all generated yml files, not sure if required to add.
We've about 115 api endpoints and each file range in 25 lines.


idlip:
I think I'll keep this as record to populate api testing files.

This would make sense if we had multiple dev or many contributors (which we dont rn)
so by time I'll try to add some security related testing file for each endpoint and add unit test there and try integrating CI checks.

But imo this would be better off in pythonic frappe tests case itself. Since we already have bunch of test suites, we can integrate for each api tests. 

In fact I had plan to populate api.md file for docs that can let user know what all they can use themselves to get data (esp useful for organizers if they want)

PR #1581, opened 2026-05-07
feat(maintainers-may): landing page for maintainers may

idlip:
image

- This will form base for future /maintainers pages and build suite of pages in similar theme
- Have been thinking of if we need a css framework and concept of reducing css to have less maintenance.

- /maintainers-may landing page
- shows events matching maintainers in title
- Show blogs from maintainers category
- A forkore themed page

PS: High LOC is from svg files


idlip:
Enough cooking. let it go live!

PR #1583, opened 2026-05-11
refactor(tests): use factories in ticketing tests

harshtandiya:
## Summary
- Adds `maximum_tickets: 100` default to `with_paid_tickets` trait on `FOSSChapterEventFactory`
- Refactors `test_foss_event_ticket.py` to use `FOSSChapterFactory`, `FOSSChapterEventFactory`, `FOSSEventTicketFactory`, `UserFactory` (drops `insert_test_chapter/event/ticket` helpers, removes hardcoded test emails)
- Refactors `test_foss_event_ticket_transfer.py` to use factories

Part of the ongoing migration from `tests/utils.py` `insert_*` helpers to the factory bot pattern (follow-up to #1560).

## Test plan
- [x] `bench --site fossunited.localhost run-tests --module fossunited.ticketing.doctype.foss_event_ticket.test_foss_event_ticket` (8/8)
- [x] `bench --site fossunited.localhost run-tests --module fossunited.ticketing.doctype.foss_event_ticket_transfer.test_foss_event_ticket_transfer` (3/3)
- [x] Regression check on existing `with_paid_tickets` users: `bench --site fossunited.localhost run-tests --module fossunited.payments.doctype.razorpay_payment.test_razorpay_payment` (12/12)
- [x] basedpyright clean on touched files

### DIFF
--- a/fossunited/tests/factories/foss_chapter_event_factory.py
+++ b/fossunited/tests/factories/foss_chapter_event_factory.py
@@ -56,6 +56,7 @@ class FOSSChapterEventFactory(BaseFactory[FOSSChapterEvent]):
                     "enabled": 1,
                     "title": "General",
                     "price": 100,
+                    "maximum_tickets": 100,
                 }
             ],
         }
--- a/fossunited/ticketing/doctype/foss_event_ticket/test_foss_event_ticket.py
+++ b/fossunited/ticketing/doctype/foss_event_ticket/test_foss_event_ticket.py
@@ -1,33 +1,23 @@
 import frappe
-from faker import Faker
 from frappe.tests.utils import FrappeTestCase
 
 from fossunited.api.checkins import checkin_attendee
 from fossunited.doctype_ids import CHAPTER, EVENT, EVENT_TICKET
-from fossunited.tests.utils import (
-    insert_test_chapter,
-    insert_test_event,
-    insert_test_ticket,
+from fossunited.tests.factories import (
+    FOSSChapterEventFactory,
+    FOSSChapterFactory,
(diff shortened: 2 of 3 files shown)


idlip:
Insert ticket is also used in `fossunited/tests/test_ticket_download.py`
Can you please update there as well (thanks!)

- Also, can you please review the ticket downloading part:

https://github.com/fossunited/fossunited/blob/d2e6fbd9e5aeb589fd5b76e6532596594447dd9d/fossunited/api/tickets.py#L561-L697

I implemented with only guard that ticket IDs are only known to buyers and not publicly known, so to download ticket incase if they did not receive email I just enabled search on (id or coupon or order_id) since only buyers know them and also rate_limit is another guard (idk if via vpn they can bypass)

- Along with that we've another api which gives out all ticket details
https://github.com/fossunited/fossunited/blob/d2e6fbd9e5aeb589fd5b76e6532596594447dd9d/fossunited/api/tickets.py#L36-L49

This again only stands that Id is not known to all. Do you think we can think of better approach to avoid brute force of iteration of many random hash?
or should we simply increase doctype.name to longer hash (so its even more harder)

1. Only possible solution i thought of was to make buying login-only (just like frappe has - bwh's `buzz`), but we need user to buy in guest mode as well (team denied login-only)

LMK your thoughts to tightening the security around tickets and ticket transfer as well.
(even we've made ticket transfer page guest accessible, but i think the api does check for owner's email to approve correctly)

idlip:
I'm merging, as I'm trying to refactor and reduce 2 api and add validations to doctype controller itself

PR #1584, opened 2026-05-11
refactor(tests): use factories in user profile and volunteers tests

harshtandiya:
## Summary
- Refactors `test_foss_user_profile.py` to use `UserFactory` (drops `insert_user_profile` and inline `frappe.get_doc` User creations). `test_share_user_with_self` keeps inline create with `ignore_permissions=True` since it intentionally simulates a Guest-context signup.
- Refactors `test_volunteers.py` to use `UserFactory.create(first_name=..., last_name=...)`, `FOSSChapterFactory.create("with_members", ...)`, and `FOSSChapterEventFactory.create(...)`. Hardcoded emails (`luffy/zoro/xebec@example.com`) replaced with random factory emails, which incidentally fixes pre-existing test flakiness caused by `USER_PROFILE` rows persisting across runs.
- Switches CI from `run-parallel-tests` to `run-tests`. The volunteers tests call `get_volunteers_data`/`get_volunteers_stats`, which query the entire `FOSS Chapter Lead Team Member` table globally. Under `run-parallel-tests`, multiple workers share one site/DB, and other workers' committed rows (e.g. orphan chapters, default `_Test1` profile) leak into these aggregate queries — failing assertions like `len(result) == 4` and `active_count == 3`. Linear `run-tests` removes the race; the speed hit is acceptable for our suite size (~27 test files).

Part of the ongoing migration from `tests/utils.py` `insert_*` helpers to the factory bot pattern (follow-up to #1560, #1583).

## Test plan
- [x] `bench --site fossunited.localhost run-tests --module fossunited.foss_profiles.doctype.foss_user_profile.test_foss_user_profile` (4/4)
- [x] `bench --site fossunited.localhost run-tests --module fossunited.tests.test_volunteers` (7/7)
- [x] basedpyright clean on touched files

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,4 +100,4 @@ jobs:
           source ./env/bin/activate
           bench --site ${{ env.SITE_NAME }} set-config developer_mode 1
           bench --site ${{ env.SITE_NAME }} set-config allow_tests true
-          bench --site ${{ env.SITE_NAME }} run-parallel-tests --app fossunited
+          bench --site ${{ env.SITE_NAME }} run-tests --app fossunited
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,4 +100,4 @@ jobs:
           source ./env/bin/activate
           bench --site ${{ env.SITE_NAME }} set-config developer_mode 1
           bench --site ${{ env.SITE_NAME }} set-config allow_tests true
-          bench --site ${{ env.SITE_NAME }} run-tests --app fossunited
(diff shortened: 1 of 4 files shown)


idlip:
Wonder why CI failed tests, locally it looks fine. Will re-run to see.

- Also I noticed in cfp submission and volunteers test each one takes 3-6 secs to run
  (this has been happening from before) can we increase the speed by changing something in class.

`Ran 7 tests in 27.466s`
this was for volunteers, is that fine or should something to be done on setup class to re-use some of it?

idlip:
CI seems to fail after re-run as well.
It fails only on Cfp submission and volunteers data. Wonder why suddenly

harshtandiya:
I will investigate

PR #1592, opened 2026-05-15
feat(cfpInsights): Assign proposal to reviewers via dashboard + notify them

idlip:
- Frappe has "ToDo" doctype which is meant to assign a doctype to work on and reminds as task tracking
- We've given IndiaFOSS co-chair desk access so they can assign user (email) to proposal they want people to review
  But with desk we fall to frappe's default way of sending hard coded notification which points to desk url (/app/doctype/...) although i've made sure to provide desk for review, we will still follow dashboard mode of approach since it can solve some UX issues.

- have added sort_by option so proposals list can be sorted based on: created date (old or newest first), review counts, assigned counts, title

- Not reinventing the wheel, just reusing the code and adapting to our platform need

Workflow:

- In "Manage chapter" organizer can goto > Cfp > Insights to see proposals list
- for each proposal they can assign multiple reviewers to it (completion populated for global reviewers)
- if assigned, the reviewer will receive email notification on the same to open "reviewers dashboard" to add review
- the list will also show who all have been assigned to in avatar with border-ring color to indicate their choice

Note :: No "ToDo" is marked as completed, since we decided to keep it open and we dont really track it (we track only review table). We can close it and show in UI otherways.. but rahul denied

Edit:
[2026-06-02]
- Query for reviewers by role so completions show for all
- Added small docs to inform the same with help tooltip icon
- Assigning only shows if event is upcoming and proposal is under "Review pending" only
- Confirm dialog before removing someone, since they won't be assigned anymore

https://github.com/user-attachments/assets/55e7e036-3ad3-4ba2-a0bd-3ca4f5f1cc19

### DIFF
--- a/dashboard/components.d.ts
+++ b/dashboard/components.d.ts
@@ -14,6 +14,7 @@ declare module 'vue' {
     AllDayGridView: typeof import('./src/components/schedule/AllDayGridView.vue')['default']
     AllProposalsBanner: typeof import('./src/components/proposals/AllProposalsBanner.vue')['default']
     AllProposalsBannerImage: typeof import('./src/components/proposals/AllProposalsBannerImage.vue')['default']
+    AssignReviewers: typeof import('./src/components/event/cfp/AssignReviewers.vue')['default']
     AttendeeCard: typeof import('./src/components/tickets/AttendeeCard.vue')['default']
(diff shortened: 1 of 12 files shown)


idlip:
I'm merging this to keep it live for indiafoss as I've tested this enough, and keeping it stale will only add more questions.
Get it live and then figure out otherwise.

This is general purpose to notify reviewers via assignment.

PR #1595, opened 2026-05-19
feat(indiafoss): IF 2026 page with image assets

idlip:
- Decided to do Indiafoss page in codebase itself under /www/, since it'll help us to reuse and programmatically track and manage page.

- Thought to do in builder only, but personally I hate to use GUI and pain to use mouse (yes!)

- Managing css is a pain, but overall for all page we need better solution to tackle that.

Have co-authored jeswin for all image assets commits (no need to thank :) )

- event doctype has new JSOn field defined as `event_data` (can be called event_meta as well)

this is to add arbitrary dynamic values from desk so page gets updated based on it.
eg: we've defined timeline so deadline varies sometimes so we simply input it in JSON field here

No more disclosure all its left is to see it live on page:
I've replicated and test 1-1 as much and should be fine to push prod now! (yes right now)

https://fossunited.org/indiafoss/2026


idlip:
I'm still not sure if we really need it in codebase under /www/
We could have done it in frappe "Web page" or builder or by templates in desk only.

But i went ahead for this time (2026) since fosshack page was also done that way, and hopefully this gives idea to optimize code so we reuse its parts in other pages (indiafoss archive when it arrives)
- or if html and css are redundant and better off to stay away from Version control then we can plan to move them out to normal "Web page"

Issue #1614, opened 2026-06-09
Send reminder on event before start date for rsvp attendees


idlip:
*This applies only for RSVP events, paid events are anyway handled by volunteers to send bulk emails*

- We already send details about event once they rsvp, but its easy to miss when the event is happening and so.
- since many are early rsvp participants, and thanks to todays busy life, we are prone to move on until we really choose the event by-heart

- Easy solution is, for rsvp we already add each participant into email group (subscription), so we can simply schedule to send notify that event is happening tomorrow or day after and inform more details if they need to better know (or attract them to attend)

- This can increase the show up count (hopefully) or keep attendees better informed.

Buzz platform send the email via custom template in codebase: https://github.com/BuildWithHussain/buzz/blob/main/buzz/templates/emails/ticket.html

We can do this, i had thought about this since it can add to more flexibility to attach calendar and so on.
or should figure out if somehow can do dynamic attachment per event.

TLDR:

Anyway the point being, send reminder for attendees before event so they show up and **never** miss.
But this might be unnecessary or adds email noise. Marketing to be called to take care ;)

chinmaychahar:
I can take this up and try to come up with something! Would that work? @idlip

idlip:
Hey @chinmaychahar ,

Yes, please feel free to contribute.

FYI we've generic email that goes to attendees when they rsvp and is accepted.
This is done by frappe's notification on new doctype creation.

<details>
<summary>Html template</summary>

```html
<html>

{% set event = frappe.get_doc("FOSS Chapter Event", doc.event) %}
{% set organiser = frappe.get_doc("FOSS Chapter", event.chapter) %}

<head>
<style>

body{

    font-family: 'Inter', sans-serif;
    width:650px;
    margin:0 auto;
    text-align: justify;
}

.mail-container{
    padding: 1.5rem !important;
}

h4{
    font-size: 1.25rem;
}

p{
    font-size: 1rem;
    font-style: normal;
    color: #1a1a1a;
    line-height: 135%;
}

a{
    text-decoration: none;
    color: #39C572;
    font-size: 1rem;
}

li{
    font-size: 1rem;
    margin: 10px 0;
}

li > span{
    color: #1a1a1a;
    font-weight: 600;
}

    @media only screen and (max-width: 620px) {
        body{
            width: 100%;
            margin: 0 auto;
        }

        .mail-container{
            padding: 1rem !important;
        }

        h4{
            font-size: 1rem;
        }

[...]

PR #1623, opened 2026-06-27
feat: send reminder email to RSVP attendees one day before event

chinmaychahar:
## Status

  - [] Draft
  - [x] Ready for review

  ---

  ## Related Issue

Closes / Addresses: #1614

  ---

  ## Problem

RSVP attendees have no reminder before an event
  ---

  ## Solution

Added a daily scheduled task that sends a reminder email to all RSVP subscribers one day before a free event starts. A `reminder_sent` flag on the event prevents duplicate sends. The email includes event name, date, time, and location with a
  link to the event page

  ---

  - [x] Tested locally
  - [ ]

  ---

  ## Changelog

  feat: send reminder email to RSVP attendees one day before event

  ---

  ## Visualization

  N/A

  ---

  ## Further Ahead

### DIFF
--- a/fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.json
+++ b/fossunited/chapters/doctype/foss_chapter_event/foss_chapter_event.json
@@ -45,6 +45,7 @@
   "column_break_ae9v",
   "event_end_date",
   "feedback_sent",
+  "reminder_sent",
   "livestreaming_section",
   "livestream_link",
   "livestream_embed_link",
@@ -492,6 +493,13 @@
    "fieldname": "feedback_sent",
    "fieldtype": "Check",
    "label": "Feedback sent?",
+   "read_only": 1
+  },
+  {
+   "default": "0",
+   "fieldname": "reminder_sent",
+   "fieldtype": "Check",
+   "label": "Reminder sent?",
    "read_only": 1
   },
   {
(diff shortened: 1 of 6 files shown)


idlip:
hey @chinmaychahar 

Thanks for the PR. Seems decent as it replicated same as feedback email flow.

Some nitpick that'd be good to have:

- We can also include "Accepted Proposers" group to remind

chinmaychahar:
Sure, SG. I'll make the changes

idlip:
Seems good to go @chinmaychahar 

I might merge this sooner or later.

Just will have internal discussion on whether to have this feature or give control to event organiser so they decide if they want this auto or simply fill template for them to send.

I'm not sure how do we solve this exactly well.

- One way I was thinking was also to create campaign on "schedule send", so organiser can edit this anytime or have this as template to be sent.

but all good from your end!

idlip:
Hey Chinmay,

Sorry to inform, but unfortunately we cannot proceed with this PR as of now.

The issue is that we as platform are sending too many emails for end users and this automation might end up being bad on our platform.

- I'll wait for community and orgnising people to comment more on the issue.
- I believe instead of force sending, another way to solve this would be to simply inform with banner in dashboard saying that event is coming soon, please update/keep them informed.
- So they make use of event mailing and manually send out.

Even feedback form can be reverted along this line, so we inform only organiser to choose what they want to proceed themself manually.

Lately the team also resonates that we've too many automated emails that sent out, that are also prone to be ignore.

Thank you for taking this up though and working on similar solution.
But we will definitely keep this PR in mind and ref back when time comes and proceed further.

PR #1631, opened 2026-07-04
fix(dashboard): deduplicate session user profile API requests

### LINKED ISSUE #809: [QUESTION] Duplicate API calls

Paramveersingh-S:
## Description
This PR resolves issue #809 by eliminating duplicate network requests for the `get_session_user_profile` API endpoint.

### Root Cause
Previously, several frontend components (`Header.vue`, `HeaderWithNav.vue`, `NewAppSidebar.vue`, and `MyProfile.vue`) independently instantiated un-cached `createResource` objects on mount. When a user navigated the dashboard, these components mounted simultaneously, dispatching multiple identical requests for the user's session profile and causing unnecessary load on the Frappe backend.

### Technical Implementation
To resolve this elegantly without needing an invasive global state refactor, I leveraged Frappe UI's native resource caching, which utilizes `swrv` under the hood. 

By defining a consistent `cache: 'SessionUserProfile'` property across all `createResource` instances querying this endpoint, Frappe UI natively deduplicates the concurrent in-flight requests. This ensures that the backend is only hit once while components still benefit from isolated success hooks and reactive bindings (e.g., `MyProfile.vue`'s downstream document resource creation).

Fixes #809

### DIFF
--- a/dashboard/src/components/Header.vue
+++ b/dashboard/src/components/Header.vue
@@ -68,6 +68,7 @@ const session = inject('$session')
 
 const user_profile = createResource({
   url: 'fossunited.api.dashboard.get_session_user_profile',
+  cache: 'SessionUserProfile',
 })
 
 if (session.isLoggedIn && session.user != 'Guest' && session.user != 'Administrator') {
--- a/dashboard/src/components/Header.vue
+++ b/dashboard/src/components/Header.vue
@@ -61,19 +61,15 @@
 </template>
 <script setup>
 import { inject } from 'vue'
-import { Avatar, Dropdown, createResource } from 'frappe-ui'
+import { Avatar, Dropdown } from 'frappe-ui'
 import FossUnitedLogo from '@/components/FossUnitedLogo.vue'
+import { fetchSessionProfile, sessionProfileResource } from '@/data/session'
 
 const session = inject('$session')
 
-const user_profile = createResource({
-  url: 'fossunited.api.dashboard.get_session_user_profile',
-  cache: 'SessionUserProfile',
-})
+const user_profile = sessionProfileResource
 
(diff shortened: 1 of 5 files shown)


idlip:
Hi @Paramveersingh-S 

thanks for trying this out, I tested this locally and..

This patch increasesd request from 2 to 3, not reducing them. 

frappe-ui createResource cache (0.1.209 and main) has no swrv and no in-flight dedup; a cache hit on an auto resource calls reload(), adding a request. Order-dependent: the other mount order drops MyProfile's onSuccess and blanks the profile page.

For real fix we might need to extract session details as shared component i suppose?

idlip:
Hey @Paramveersingh-S 

Kindly pinging you again, in case you would like to update anytime you are free.

Paramveersingh-S:
Sorry for late reply @idlip 
I will update you tomorrow evening.

Paramveersingh-S:
Hey @idlip! 
You were right about the Frappe UI createResource cache lacking in-flight deduplication. Because all the components were mounting at exactly the same time, they were bypassing the cache before the first request even finished, which caused the extra network calls.
To fix this properly, I've exported a shared sessionProfileResource and a singleton fetchSessionProfile() function in session.js. It uses a module-level Promise tracker to explicitly deduplicate any in-flight requests. I also used Vue's native watch in MyProfile.vue to react to the shared data, ensuring the onSuccess hook never gets dropped regardless of mount order.

Please retest if it worked or not

idlip:
tested and works with one call.

thanks for the fix!

Paramveersingh-S:
That's great

Issue #1630, opened 2026-07-01
bug: speaker social link breaks for some speakers


exitflynn:
[Some](https://fossunited.org/c/indiafoss/2026/cfp/f2cko4si04) [of](https://fossunited.org/c/indiafoss/2026/cfp/9a8056ifdl) the speakers' social links in IndiaFOSS CFPs are broken while some others' work [fine](https://fossunited.org/c/indiafoss/2026/cfp/1crh77ogup).

Looks like this happens when a speaker saves their link without https:// (e.g. `www.linkedin.com/..` or `linkedin.com/...`) and the browser treats it as relative and resolves it against the current page, producing a broken URL like https://fossunited.org/c/indiafoss/2026/cfp/www.linkedin.com/in/peter-parker-673624184/)

idlip:
Ah thanks for reporting!

They are missing https prefix.
We had added http helper for profile page, now is good time to cover for all other missing pages/fields.

idlip:
Have fixed all social links form from backend.

Ideally adding URL field option validates it, for frontend UX the ensure http helps to prefix this.

Fixed (hopefully everywhere)!

PR #1634, opened 2026-07-08
fix(cfp): normalize speaker social link href on the frontend to include https:// scheme

exitflynn:
Speaker social links saved without a URL scheme (e.g. www.linkedin.com/...) were rendered as relative hrefs on the CFP proposal page, resolving against fossunited.org instead of the intended external site.

## Status

- [ ] Draft
- [x] Ready for review

---

## Related Issue

Addresses: #1630 

---

## Problem

Speakers' profile links without https:// would be rendered to relative links, resolving against fossunited.org.

---

## Solution

One fix for [this](https://github.com/fossunited/fossunited/issues/1630#issuecomment-4864456175) was already done on the backend by updating the URLs in db I think. This is another way to handle this which I'd discovered during reporting, where we add handling in how the frontend handles url's, the codebase has a [similar](https://github.com/fossunited/fossunited/blob/0c080d20833c411e8ab749d3c6f7b1d40cc25158/dashboard/src/helpers/utils.js#L43-L46) handling in other places.

This PR was super easy to open since I just had this pushed and lying around and felt might as well get eyes on this for discussions haha. The outcome of closing this in favour of the current fix is totally 🆒 .

---

- [x] Tested locally
- [ ]

### DIFF
--- a/fossunited/fossunited/doctype/foss_event_cfp_submission/templates/foss_event_cfp_submission.html
+++ b/fossunited/fossunited/doctype/foss_event_cfp_submission/templates/foss_event_cfp_submission.html
@@ -261,8 +261,9 @@
     />
   {% if sp_url %}</a>{% else %}</div>{% endif %}
   {% if speaker.social_link %}
+    {% set social_url = speaker.social_link if speaker.social_link.startswith('http://') or speaker.social_link.startswith('https://') else 'https://' + speaker.social_link %}
     <a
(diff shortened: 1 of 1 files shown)


idlip:
Oh cool. Thanks for having this.
Ideally I dont think we require, since I changed every social_link field in backend once and fundamentally we've URL validation from frappe itself.

Anyway would not hurt to have it for edgecase.

PR #1635, opened 2026-07-09
feat(indiafoss): archive page to aggregate all indiafoss talks

idlip:
IndiaFOSS Archive will kinda mimic "Mini Youtube"
Basically we are showing youtube videos itself, just that with speaker names and proposal page link. Aggregation is the main focus!

- `/indiafoss/archive` page acts as aggregate to all previous Talks to show
  - Click on thumbnail plays the video in-page
  - Click on title takes to proposal page (for IndiaOS (1), 2.0 it takes to YT link itself, since there is no CFP page in archive platform)

- Doctype :: `Event Media` which collects YT URL and links each to an proposal (so auto fetch other details)
  - Helps to also map older events and is generic to expand for other events to show in some page like `/videos` or `/media`

- Data migration is hard and will be incomplete, and we might miss the taste and state of how platform was that time, so IMHO it better stay legacy on its own page.

- Further additions `/indiafoss/speakers` to show all aggregate of speakers who spoke in IndiaFOSS and each speaker page shows talks list as well.

TBH IMO its not required for platform, since it does reinvent the wheel and added new way to show. But I thought it can stay generic to collect for all events (Mumbai, MangaloreFOSS and etc) so we group and once place rules them all.

image

### DIFF
--- a/fossunited/doctype_ids.py
+++ b/fossunited/doctype_ids.py
@@ -30,6 +30,7 @@ CORE_TEAM = "Core Team Member"
 EVENT_VOLUNTEER = "FOSS Chapter Event Member"
 EVENT_SCHEDULE = "FOSS Event Schedule"
 COMMUNITY_PARTNER = "FOSS Event Community Partner"
+EVENT_MEDIA = "Event Media"
 
 # Event proposal-related identifiers
 EVENT_CFP = "FOSS Event CFP"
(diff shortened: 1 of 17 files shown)


idlip:
definitely came to too many LOC.
Has come gaps to refactor back.

But for now serves its purpose and seems to be working as intended.

All is there to merge and on prod add doctype content for all to map properly.

PR #1639, opened 2026-07-15
fix: Event photos ente link in clubs-peq.md 

siddharth2798:
## Status

- [ ] Draft
- [x] Ready for review


idlip:
@siddharth2798 you still do not have access to self merge? Adding it to fossunited org group should've given. 
Anyway no worries!

Issue #1662, opened 2026-07-29
Bump all python deps in pyproject.toml


idlip:
- Make sure to test and bump each package

- With setup tools removing pkg_resources, its annoying that old razorpay depends on it.

Gotta keep updated to remove issues like this (made me debug on prod today)

idlip:
Tested with new bench instance and from scratch install.

Even seed data works fine (except for show_rsvp being deprecated, will remove in next PR)

But version bump should be fine from our end.

Note: Frappe also bunldes PyGithub and Pillow, so only package we gotta have in pyproject is Razorpay for payment, if not everything is bundles from frappe itself

PR #1665, opened 2026-07-30
build: bump razorpay + other python package to latest

idlip:
- Fix on #1662 

- Especially for razorpay to get rid of using pkg_resources from setuptools

- Other stuff should resolve fine. test with razorpay payment, and it did not have much changes


idlip:
Tested with new bench instance and from scratch install.

Even seed data works fine (except for show_rsvp being deprecated, will remove in next PR)

But version bump should be fine from our end.

Note: Frappe also bunldes PyGithub and Pillow, so only package we gotta have in pyproject is Razorpay for payment, if not everything is bundles from frappe itself

PR #1589, opened 2026-05-13
migrate: frappe version v15 -> v16

idlip:
- Migrate from v15 -> v16

https://github.com/frappe/frappe/wiki/Migrating-to-version-16

- [x] Make blog & newsletter required apps
- [x] Adjust all get_all to [[https://github.com/frappe/frappe/wiki/query-builder-migration][query builder migration]]
- [x]  see if `db.get_value` casts single doctype results affects anywhere
- [x] Update CI/CD for that version (check if boilerplate has any update to do)
- [ ] Update development docs
- [ ] Check for PDF output since frappe v16 switches PDF rendering from wkhtmltopdf -> Chromium

- In cloud try with new bench separately to see if everything works then take site backup and then move.

Lot of things to cross-verify and so. I'm only opening this to kickstart the work.

Was annoyed by facing `pkg_resources` not found error, had to go venv and install setuptools and then get-app install-app... idk where to blame but annoys!

PR #1684, opened 2026-08-07
fix(grants): fix currency option to get sum on number card

idlip:
replaced option field: `INR` -> `Company:company:default_currency`

- Issue in upstream repo that does not allow using sum on number card for currency

- this fix should resolve it: https://github.com/frappe/frappe/issues/18611#issuecomment-3761068367

- also will not be required & can revert back to INR in frappe version-16

- [X] Test locally to have number card and it shows fine in INR itself.

Siddharth wants it to show on workspace (/app) so its easy to get total disbursed in timeframe.


idlip:
All good, no concern I guess.
No need company field or anything i think as per above llm warning.

idlip:
backlink #1589

PR #1692, opened 2026-08-13
chore: add mock data for IndiaFOSS 2026 to seed script

jasilfaras:
The local development environment for the IndiaFOSS 2026 page was missing a lot of data, which caused several UI components (like the timeline, sponsors, devrooms, and FAQs) to not render at all. 

This PR adds a `seed_indiafoss_2026()` function to populate the database with mock data. Getting these components to render properly locally should help developers test UI changes and make it easier to contribute accessibility improvements. I tried to get the layout as close to the real website as possible without adding the overhead of handling custom assets (like cover images or actual sponsor logos).

### Changes
* Added `seed_indiafoss_2026()` to generate mock events, sponsors, devrooms, and FAQs.
* Added specific `teardown_all()` logic to safely clean up this mock data.

### Screenshots

**Before:**

Before
Before

**After:**
After
After
After
After

### How to test
```bash
bench execute fossunited.dev.seed.seed_indiafoss_2026
```

### DIFF
--- a/development/seed.py
+++ b/development/seed.py
@@ -7,6 +7,7 @@ Usage::
 
     # For Frappe Manager / Manual Bench installations:
     bench execute fossunited.dev.seed.seed
+    bench execute fossunited.dev.seed.seed_indiafoss_2026
     bench execute fossunited.dev.seed.teardown_all
 
     # For Devcontainer (Docker/Podman) environments:
@@ -27,7 +28,9 @@ Lead       {bangalore,mumbai,kochi,campus}-lead@...      password
 Idempotent — safe to run repeatedly on the same database.
 """
 
+import json
 import random
+from contextlib import contextmanager
 from datetime import datetime, timedelta
 
 import frappe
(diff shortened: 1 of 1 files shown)


idlip:
Hey @jasilfaras ,

Thanks for taking up to add this.
But I wanted to keep seed data very concise, in fact my future plan was to refactor more to make it as a tool that is more deterministic on what gets added. (maybe tui/fzf asking questions to add and remove)

IMO script is already blown up, I do not want to add indiafoss data here.
Although its good for contributions, wouldn't it be better to just add event and the json data can be simply in a file for them to copy paste?

tbh even this might turn our redundant since I dont know how many really did setup and has this going, they can always message me or raise issue here so i can help for quick setup.
But yea, seed data is meant for long term that anytime it can be run. Indiafoss 2026 might be temporary.

https://fossunited.org/files/indiafoss-26.json

^ here's the json data that we use for "event_data" field for indiafoss 2026.

jasilfaras:
Ah got it didn't realize it would be easier to just paste the json directly from the UI.ill go ahead and close this pr

PR #1697, opened 2026-08-15
fix: ambiguous link text (here/apply/Read More) across docs and web pages

### LINKED ISSUE #1689: [A11y] Audit for "Read more" or "click here" link ambiguity
- Although making a link with text as "read more" or check "here" is kinda clean for normal readers, it still is a bad practice to put it out in wild text.

Even disrupts quick tab jump that means nothing out of context.

- With some rich-text-editor or markdown we do not have control to add custom aria-label, so its better to avoid this short text itself and use full form to inform what it is.

- This is not a specific page issue, this is to be covered across all pages ("Web Page" and custom desk ones as well)

saksham242007:
Closes #1689

Problem
The platform used ambiguous link text like "here", "apply", and "Read More" in several docs and web pages. Screen reader users tabbing through links or using a links-list view can't tell where these links go without reading the surrounding paragraph.

Solution

Audited the repo for ambiguous link text ("here", "click here", "read more", "apply", generic verbs used as link text) across HTML, Vue, and Markdown files. Replaced 10 instances with descriptive link text so each link's purpose is clear out of context. No new ARIA attributes added — fixed via visible link text per the issue's guidance.

### DIFF
--- a/docs/docs/clubs-contribution-ideas.md
+++ b/docs/docs/clubs-contribution-ideas.md
@@ -74,6 +74,6 @@ If you have other ideas, feel free to book a call to discuss with the foundation
 
 
 
-> Have other ideas to suggest? Send us a PR [here](https://github.com/fossunited/fossunited/tree/develop/docs/docs).
+> Have other ideas to suggest? Send us a [PR with your idea](https://github.com/fossunited/fossunited/tree/develop/docs/docs).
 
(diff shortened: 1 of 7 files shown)


idlip:
Thank you for correcting them.
lol even I did not think we had few like this.

PR #1698, opened 2026-08-15
fix(tickets): revalidate orders before fulfillment

anxkhn:
## Summary

Revalidate paid ticket orders immediately before fulfillment, serialize paid capacity allocation, and automatically refund captured payments that can no longer be fulfilled.

## Context

I found this while researching a technical write-up about FOSS United's open-source Razorpay integration. A Razorpay Order created while a ticket tier is valid remains payable after the tier expires or sells out because Razorpay Orders do not expire automatically.

I am submitting the remediation before publishing the write-up. The write-up will not include live payment identifiers.

## Problem

Ticket tier validity and capacity are checked when the pending `Razorpay Payment` record is created. Between order creation and capture, however:

- the event can close ticket sales;
- a tier can be disabled or pass `valid_till`;
- other buyers can consume the remaining capacity;
- callback and webhook delivery can race or replay.

The current capture path marks the payment `Captured` and creates tickets without revalidating those conditions. Capacity checks also use a count without serializing concurrent paid fulfillments, so two captures can observe the same last available seat.

## Changes

- Route browser callbacks, Razorpay webhooks, and manual payment synchronization through a locked, monotonic `capture_payment` transition.
- Lock the parent event row before checking existing fulfillment, tier validity, and current capacity, keeping validation and ticket insertion in one transaction.
- Treat callback/webhook replay as an idempotent no-op once a payment leaves `Pending`/`Failed`.
- Move captured but unfulfillable payments to `Refund Pending` without creating tickets.
- Enqueue the refund only after the database transaction commits.
- Use Razorpay's `X-Refund-Idempotency` header so a lost response can be retried safely.
- Add an hourly recovery task for refund intents where no Razorpay refund ID was recorded.
- Preserve late-authorized payment handling by permitting `Failed -> Captured`.

## Why the event row is locked

A count followed by inserts is subject to a time-of-check/time-of-use race. Locking the event row serializes paid ticket fulfillment for that event, so the second capture observes tickets committed by the first before deciding whether capacity remains. The lock also serializes callback/webhook processing around the existing payment-level lock.

## Refund behavior

Razorpay Orders do not support a server-side expiry timestamp. If a previously valid order is paid after it can no longer be fulfilled, the customer has already been charged. The patch therefore queues a full refund rather than leaving the customer to request one manually.

The refund worker uses a stable idempotency key derived from the local payment record. The scheduler retries only `Refund Pending` records without a stored Razorpay refund ID; normal `refund.processed` webhooks continue to finalize refund state.

## Tests

Added focused coverage for:

- capture after tier expiry;
- capture after tier capacity is exhausted;
- callback/capture replay;
- late capture after an earlier failed state;
- stable refund idempotency keys.

## Validation

- `ruff check` on all changed Python files
- `ruff format --check` on all changed Python files
- project pre-commit hooks on all changed files
- `python -m compileall -q fossunited`
- `git diff --check`

A full Frappe integration test run requires a configured Bench site and was not available in this checkout.

### DIFF
--- a/fossunited/hooks.py
+++ b/fossunited/hooks.py
@@ -165,6 +165,9 @@ before_migrate = "fossunited.migrate.before_migrate"
 # Scheduler Events
 # ----------------
 scheduler_events = {
+    "hourly": [
+        "fossunited.payments.doctype.razorpay_payment.razorpay_payment.retry_pending_refunds",
+    ],
     "daily_long": [
         "fossunited.scheduled_tasks.conclude_events",
     ],
--- a/fossunited/hooks.py
+++ b/fossunited/hooks.py
@@ -165,9 +165,6 @@ before_migrate = "fossunited.migrate.before_migrate"
 # Scheduler Events
 # ----------------
 scheduler_events = {
-    "hourly": [
-        "fossunited.payments.doctype.razorpay_payment.razorpay_payment.retry_pending_refunds",
-    ],
     "daily_long": [
         "fossunited.scheduled_tasks.conclude_events",
     ],
--- a/fossunited/api/dashboard.py
+++ b/fossunited/api/dashboard.py
@@ -9,6 +9,7 @@ from fossunited.doctype_ids import (
     TICKET_TIER,
     USER_PROFILE,
 )
+from fossunited.payments.doctype.razorpay_payment.razorpay_payment import capture_payment
 from fossunited.utils.payments import (
     get_in_razorpay_money,
     get_razorpay_client,
@@ -157,11 +158,7 @@ def handle_payment_success(order_id: str, payment_id: str, signature: str):
         }
     )
 
-    # update payment
-    payment = frappe.get_doc(RAZORPAY_PAYMENT, {"order_id": order_id})
-    payment.status = "Captured"
-    payment.payment_id = payment_id
-    payment.save(ignore_permissions=True)
+    capture_payment(order_id, payment_id)
 
 
 # nosemgrep: guest-whitelisted-method
--- a/fossunited/api/dashboard.py
+++ b/fossunited/api/dashboard.py
@@ -158,7 +158,7 @@ def handle_payment_success(order_id: str, payment_id: str, signature: str):
         }
     )
 
-    capture_payment(order_id, payment_id)
+    return {"status": capture_payment(order_id, payment_id)}
 
 
 # nosemgrep: guest-whitelisted-method
--- a/fossunited/api/dashboard.py
+++ b/fossunited/api/dashboard.py
@@ -1,3 +1,5 @@
+import time
+
 import frappe
 from frappe import _
 
@@ -121,6 +123,7 @@ def create_razorpay_order(
         data={
             "amount": get_in_razorpay_money(amount),
             "currency": "INR",
+            "expire_by": int(time.time()) + 30 * 60,
         }
     )
 
--- a/fossunited/api/dashboard.py
+++ b/fossunited/api/dashboard.py
@@ -1,5 +1,3 @@
-import time
-
 import frappe
 from frappe import _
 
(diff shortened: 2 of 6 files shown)


anxkhn:
hey @idlip, can you have a look if available?

fossunited/hooks.py
```diff
@@ -165,6 +165,9 @@
 # Scheduler Events
 # ----------------
 scheduler_events = {
+    "hourly": [
+        "fossunited.payments.doctype.razorpay_payment.razorpay_payment.retry_pending_refunds",
+    ],
     "daily_long": [
```
idlip, line 168: Hey @anxkhn 

We don't really do refunds, thus we have strict policy of saying only transfers are allowed.

https://fossunited.org/refund-transfer-policy

So can you please remove this scheduler here and other refund parts of code.
Thanks for opening this PR, I'll review and comment further later

idlip:
Hey @anxkhn 

Thanks for digging into this.

Verified core problem: Razorpay orders don't expire by default, so an order created before a tier sells out, gets disabled, or expires can still be captured later with no recheck. The revalidation at capture time and the event-row lock (to stop two payments both grabbing the last seat) are both changes we want to keep.
  
Before merging, please make these changes:

1. Drop the refund automation entirely: 
process_refund, retry_pending_refunds, the hourly scheduler hook, and the Refund Pending status transition. We run a [strict no-refund, transfer-only policy](https://fossunited.org/refund-transfer-policy), so this path will never be used. When validate_ticket_fulfillment fails after capture, just log the error (`frappe.log_error`) and leave the payment as Captured with no ticket. We'll handle those cases manually (the existing refund() method on the doctype already covers it if we ever decide to use it).

2. Please also set `expire_by` when creating the Razorpay order in `create_razorpay_order` (fossunited/api/dashboard.py), to a flat 30mins from creation. Right now we don't set it at all, so orders stay payable forever (in backend). A short fixed expiry clears stale/duplicate payment attempts.

3. Keep the fix to client.payment.refund(self.payment_id, {"amount": refund_amount}) in the existing manual refund() method. That's a separate real bug (wrong call signature), unrelated to the no-refund policy, and worth keeping.

Once the refund automation is removed and expire_by is added, this looks good to merge.

anxkhn:
hey @idlip, dropped the refund automation and hourly scheduler. failed fulfillment now stays captured with no ticket and only logs the error. also set expire_by to 30 minutes on create_razorpay_order, and kept the refund() payload fix.

note: razorpay's current orders create docs still don't list expire_by (that's a payment-link field). i set it as requested; if live order create starts rejecting the field, we can drop it and keep the capture revalidation only.

anxkhn:
hey @idlip, one follow-up after double-checking the full flow and Razorpay’s current documentation.

`expire_by` is supported for Payment Links, but Razorpay explicitly says it is rejected for Orders + Standard Checkout. Their `timeout` checkout option only closes that modal; it does not invalidate the underlying Order, which can still be paid later.

A 30-minute application-side cutoff at capture would not prevent payment either. It could leave a customer charged with no ticket even when the tier remains available, which is worse under the no-refund policy.

My recommendation is therefore to:

- remove `expire_by` from the Order payload;
- optionally add Checkout `timeout: 1800` as a UX guard only;
- keep the capture-time event/tier/capacity revalidation and event-row lock, so stale orders are fulfilled only when they remain valid and capacity is available;
- fix the remaining race where a failure callback can overwrite a successful `Captured` transition.

I also verified that fulfillment should happen only after Razorpay confirms the payment is `captured`; signature verification proves authenticity but not capture status.

Does this approach work for you? I have not changed the expiry behavior again pending your confirmation.

idlip:
Sorry my bad, thought there would be easy option like this. 

Seems like there is pending issue: https://github.com/razorpay/razorpay-node/issues/426

This can be kept sep issue, cause we've lot of payment log that is stalled as "Pending"
I think at some point its better to clear things up.

You can remove expire_by field. 

So far good, I'll try to test this locally once.
Just curious, Were you able to setup our platform and test this locally?

anxkhn:
yeah, i was able to set up the platform locally and run the focused frappe integration tests. all 20 razorpay payment tests passed.

removed `expire_by` as suggested and also fixed the failed/captured transition race. pushed in `c28fe8d1`.

agree the stalled `pending` payment logs are better handled as a separate cleanup/reconciliation issue.

idlip:
Hey thanks anas khan, 
It took some time for me to sit down and test.

With main fix, only concern is edgecase scenario, where user went into 4th step and we disabled ticket or time is over, at this point when they make payment it gets captured and ticket wont get generated.

I think might be fine since you made it to error log and if they reach by email we can resolve.
Just hope it does not cost us much!

Thank you self evaluating and fixing a gap for us!
Much appreciated.

PR #1704, opened 2026-08-19
fix(security): harden ticket & checkin endpoints

sidd190:
## Summary

From reference to the gist summary.

- **VULN-002 (HIGH):** `get_ticket_details` returned `["*"]` (28 fields) to guests -- replaced with 8 explicit safe fields. Removes exposure of email, payment details, and internal metadata. Addresses [GHSA-8w2h-86fx-99fc](https://github.com/fossunited/fossunited/security/advisories/GHSA-8w2h-86fx-99fc).
- **VULN-007 (MED):** Stored XSS via `| safe` in IndiaFOSS speaker talks template -- now sanitized with `frappe.utils.sanitize_html()`.
- **VULN-008 (MED):** `get_tickets_insights`, `get_checkin_insights`, `get_ticket_tiers` had no rate limits -- added `@rate_limit(limit=10, seconds=3600)`.
- **Additional hardening:**
  - `create_transfer_request` returns `{name, status}` instead of full doc (was leaking receiver_email, owner_email, Frappe metadata).
  - `get_tickets_insights` tier query `["*"]` replaced with explicit fields.
  - `get_attendee_with_checkin_data` filter keys whitelisted -- previously any column could be injected via the `filters` dict. Also fixes a pre-existing crash when `filters=None`.

Bruno security test assertions added to all 20 Group 1 endpoint files (api/tickets + api/checkins).

## Test plan

- [x] All 7 tests in `test_ticket_download.py` pass
- [x] All 5 tests in `test_speaker_coupons.py` pass
- [x] Full suite (249 tests): 0 new failures
- [x] Frontend regression check: `TicketTransfer.vue`, `PaymentSuccess.vue`, `EventCheckins.vue`, `QuickEventCheckin.vue`, `get_event_tickets.html` -- all field dependencies verified
- [ ] Live test: `get_ticket_details` as guest excludes sensitive fields
- [ ] Live test: ticket transfer, search, checkin flows work end-to-end


idlip:
Hey @sidd190 

can you please keep bruno test files to diff PR, these are too many.
I only thought you were adding for get_ticket_details to verify.

Please do atomic PR, let it cover only for necessary parts.
Whole bruno files are separate and this is introducing too many changes to track.

you can only keep: bruno-collection/api/tickets/get-transfer-details.yml here for now

idlip:
> VULN-008 (MED): get_tickets_insights, get_checkin_insights, get_ticket_tiers had no rate limits -- added @rate_limit(limit=10, seconds=3600).

Also no need these, these are sometimes (during event) meant to be refetched many times.

Infact its all counts data here, and safe for guest users to get as well. You can remove rate_limit or keep time span very less.

idlip:
LGTM.

Thanks.
Merging as its minimal and tested.

Please update docs for bruno tests and add screenshots for same verification of tests!

PR #1711, opened 2026-08-20
fix(css): untangle interleaved jobs and clubs list-view styles

sidd190:
## Summary

- The merge of #1664 (jobs) after #1663 (clubs) left the CSS blocks spliced into each other
- `.v3-job-list-header` and `.v3-job-list-row` were never closed with `}`, breaking both list views
- `.v3-club-list-row:hover` was duplicated
- This separates the two blocks so each is self-contained and properly closed

## What was broken

1. `.v3-job-list-header {` opened but never closed -- clubs comment + rules injected mid-declaration
2. `.v3-job-list-row {` opened but never closed -- `.v3-club-list-row` nested inside it
3. Duplicate `.v3-club-list-row:hover` rule


idlip:
Thank you for the fix!

Git Web conflict editor sucks

PR #1712, opened 2026-08-21
feat(free-ticket): ask t-shirt size if coupon has t-shirt included

### LINKED ISSUE #1706: Collect T-shirt size from free ticket applications
- for IndiaFOSS 2026 we might give out t-shirts to speakers and volunteers, so we need a way to ask for size while they take a coupon via web form

- Also give option in code to say if t-shirt is included or not, so we organiser can also control it.

- needs fix on #1554 as well since it shows only count for auto sending script

idlip:
- Fix #1706 

- Most paid events might give out t-shirts to Speakers and volunteers.

When we give free coupon we fail to collect t-shirt size from them to order properly.
So coupon gets checkbox if t-shirt included and application ask with select option

- Web form will only show this option if coupon includes t-shirt. Thus had to introduce an api which simply says true or false for t-shirt

- We've controller to avoid anyone entering t-shirt even if coupon does not have.

- Dashboard create coupon showing t-shirt checkbox
image

### DIFF
--- a/dashboard/src/components/event/FreeTicketCodeDialog.vue
+++ b/dashboard/src/components/event/FreeTicketCodeDialog.vue
@@ -45,6 +45,12 @@
           :min="1"
           :max="30"
         />
+        <FormControl
+          v-model="code.tshirt_included"
+          type="checkbox"
+          label="T-shirt included"
+          description="Ask for a t-shirt size when this coupon is claimed"
+        />
       </div>
       <ErrorMessage :message="errorMessages" />
     </template>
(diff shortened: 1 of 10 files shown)


idlip:
One missing piece is, asking arbitrary custom field questions.

Normal ticket can hold some question like gender or food, so we miss collecting them.

We can add same table in application and make web form help fill question to collect response.

idlip:
Have tested and should be good to go live.

TLDR

- Free ticket code gets checkbox to ask if tshirt can be included

- Application asks for t-shirt size conditionally and ticket reflects the same as filled

- Additionally, added custom field table so, any ticketing question asked reflect in web form (via client script). To complete the flow

PR #1722, opened 2026-08-27
fix(security): restrict hackathon endpoint field exposure

sidd190:
## Summary

The `get_hackathon`, `get_hackathon_from_permalink`, `get_participant`,
and `get_project_by_team` endpoints were using `frappe.get_doc()` which
returns every field on the document, including internal/admin fields.

This replaces all four with `frappe.db.get_value()` calls using explicit
`SAFE_FIELDS` lists so only the fields the frontend actually needs are
returned.

### What changed

- `fossunited/api/hackathon.py`:
  - `get_hackathon()` - SAFE_FIELDS allowlist (17 fields)
  - `get_hackathon_from_permalink()` - same SAFE_FIELDS allowlist
  - `get_participant()` - explicit field list (11 fields)
  - `get_project_by_team()` - PROJECT_FIELDS allowlist (12 fields)

### Bruno tests added (3 tests)

- `01-get-hackathon-safe-fields.bru` - verifies response contains only
  allowed fields and no internal fields like `_user_tags` or `owner`
- `02-get-hackathon-from-permalink-safe-fields.bru` - same check via
  permalink lookup
- `03-get-hackathon-nonexistent.bru` - verifies proper error for invalid
  hackathon ID

### Merge sequence

Merge after: `chore/precommit-bruno-hook`
Merge before: `fix/security-group-3-cfp-forms`

## Test plan

- [ ] Open a hackathon page on the frontend and verify it still loads
  correctly with the restricted field set
- [ ] `cd bruno-collection && npx @usebruno/cli run api/hackathon --env local-development`
  (3 passed, 0 failed)


fossunited/api/hackathon.py
```diff
@@ -119,9 +163,24 @@ def get_participant(hackathon: str) -> dict:
...
+            "is_student",
+            "git_profile",
+            "organization",
+            "hackathon",
+            "wants_to_attend_locally",
+            "localhost",
+            "localhost_request_status",
+        ],
+        as_dict=True,
     )
```
idlip, line 184: Do we really need to guard here?

Since the participant is the owner itself, they can get everything right?
since the guard is only on session.user?

to check for more permission with get_doc we can use:

`doc.check_permission("read")` so that should be good?

Actually hackathon segments itself needs more review.

Please skip hackathon code stuffs to after indiafoss. For now tickets and events are most important.

sidd190: Ohhh okayy! There were just so many fields under session.user which felt unnecessary/extra to return back to the user, and it's kinda better to keep the fields and stuff to ourselves where we don't need to return them, but should I convert it to draft for now?

fossunited/api/hackathon.py
idlip, line 1: As per #1737 Please move both the `SAFE_FIELDS` to its respective doctype ("FOSS Hackathon") place so its reusable.

Same for participant and Project fields.

idlip (changes requested):

idlip:
@sidd190 

tests pass and is good to go.

as per #1737 , please move both SAFE_FIELDS, hackathon participant fields and project fields to its own respective doctype?

PR #1725, opened 2026-08-27
fix(security): fix RSVP mass assignment, path traversal, and checkin auth

sidd190:
## Summary

Three separate issues fixed across RSVP and profile endpoints.

### What changed

1. **RSVP mass assignment** (`fossunited/chapters/doctype/foss_event_rsvp/foss_event_rsvp.py`):
   `create_rsvp()` passed the entire user-supplied `fields` dict
   directly to `frappe.get_doc()`, allowing injection of fields like
   `status`, `event`, `chapter`, or `submitted_by`. Now filters through
   an `ALLOWED_FIELDS` set and derives `event`, `event_name`, and
   `chapter` from the linked RSVP form server-side.

2. **Profile image path traversal** (`fossunited/api/profile.py`):
   `set_profile_image()` and `set_cover_image()` used
   `frappe.get_site_path("public", file_url.lstrip("/"))` which allowed
   `../../` sequences to read files outside the public directory. Added
   `_validate_file_path()` that resolves symlinks via `os.path.realpath()`
   and checks the result stays within the site's public directory.

3. **Self-checkin authorization bypass**
   (`fossunited/chapters/doctype/foss_event_rsvp_submission/foss_event_rsvp_submission.py`):
   `self_check_in()` and `remove_checkin_for_today()` only checked
   `frappe.session.user != "Guest"` but not whether the caller actually
   owns the submission. Added `doc.submitted_by != frappe.session.user`
   check so users can only check in/out their own RSVPs.

### Bruno tests added (6 tests)

- `api/rsvp/00-login-as-attacker.bru` - authenticates as a different
  user (not the RSVP owner)
- `api/rsvp/01-create-rsvp-rejects-status-injection.bru` - sends
  `status: "Confirmed"` in the payload, verifies it's stripped
- `api/rsvp/02-self-checkin-rejects-non-owner.bru` - non-owner tries to
  check in on someone else's RSVP, expects rejection

- `api/profile/00-login.bru` - authenticates
- `api/profile/01-set-profile-image-path-traversal.bru` - sends
  `../../etc/passwd` as file_url, expects rejection
- `api/profile/02-set-cover-image-path-traversal.bru` - same for cover
  image

[Screencast From 2026-09-02 17-37-28.webm](https://github.com/user-attachments/assets/454dfd32-015f-4779-90e7-139529957c8f)


fossunited/chapters/doctype/foss_event_rsvp/foss_event_rsvp.py
```diff
@@ -132,15 +132,32 @@ def check_if_already_rsvp(self):
...
     linked_rsvp_exists = frappe.db.exists(EVENT_RSVP, fields.get("linked_rsvp"))
     if not linked_rsvp_exists:
         frappe.throw(_("Invalid RSVP ID."), frappe.DoesNotExistError)
 
-    fields.update(
+    rsvp_doc = frappe.get_doc(EVENT_RSVP, fields["linked_rsvp"])
+
+    safe_fields = {k: v for k, v in fields.items() if k in ALLOWED_FIELDS}
+    safe_fields.update(
         {
```
idlip: Actually its not clean to mention Safe fields and Allow fields inside functions.

Its difficult to know what field exist or to allow in future.

Won't it be best to define that as constant in its respective doctype?

for this, rsvp_submission.py itself can have this so easy to point where its coming from and exists?

fossunited/chapters/doctype/foss_event_rsvp/foss_event_rsvp.py
```diff
@@ -132,15 +132,32 @@ def check_if_already_rsvp(self):
...
+
     fields = json.loads(fields)
 
     linked_rsvp_exists = frappe.db.exists(EVENT_RSVP, fields.get("linked_rsvp"))
     if not linked_rsvp_exists:
         frappe.throw(_("Invalid RSVP ID."), frappe.DoesNotExistError)
 
-    fields.update(
+    rsvp_doc = frappe.get_doc(EVENT_RSVP, fields["linked_rsvp"])
+
```
idlip: even this can be condensed to only pass required field as below?

event, event_name and chapter?
just small optimization

idlip:
Except those small nitpick changes above, rest LGTM.

Please attach screenshots if you ran the bruno tests with checkbox to document the PR as done.

Thanks.

ni5arga:
Hi,

> Profile image path traversal (fossunited/api/profile.py):
set_profile_image() and set_cover_image() used
frappe.get_site_path("public", file_url.lstrip("/")) which allowed
../../ sequences to read files outside the public directory. Added
_validate_file_path() that resolves symlinks via os.path.realpath()
and checks the result stays within the site's public directory.

Do you have a proof of concept for this? I do not think you can read any sensitive file via this method.

It was flagged by me before internally but it was concluded that it is not exploitable pillow crashes as sensitive files like `site_config.json` are not in a pillow-compatible (image) format.

telegram-cloud-photo-size-5-6057409948167114213-y


It will be very interesting if you can somehow read the files which have sensitive data in them.

sidd190:
> Do you have a proof of concept for this? I do not think you can read any sensitive file via this method.
> 
> It was flagged by me before internally but it was concluded that it is not exploitable pillow crashes as sensitive files like `site_config.json` are not in a pillow-compatible (image) format.
> It will be very interesting if you can somehow read the files which have sensitive data in them.

Hey @ni5arga, this is more of a hardening/defense-in-depth fix rather than direct exploitation.

The problem here can be private image exfiltration rather than direct sensitive data access. The attack would look something like,

- Frappe backup files follow a date-based naming convention (YYYY-MM-DD_HHMMSS-*-database.sql.gz). These aren't images so Pillow would reject them, but the error response differs from "file not found", creating a file existence oracle that confirms whether a backup exists at a guessed path.
- The endpoint returns different errors for "file not found" vs "not a valid image" vs "path rejected". An attacker can enumerate private/files/ contents by probing common filenames and reading the error type. This works for any file, not just images.
- When the target is an image (e.g. uploaded ID documents, private event photos), the endpoint reads it through Pillow, converts to WebP, and saves it as a public profile photo. The private image is now accessible to anyone. Frappe uses original-filename + 6-char content hash as the suffix, so filenames aren't fully random.
- The oracle lowers the bar for the attack. Enumerate until you find a .png/.jpg, then exfiltrate it.

I feel it's not critical, but worth applying the fix and hardening just in case.

PR #1726, opened 2026-08-27
fix(cfpForm): override global guidelines via cfp form description

### LINKED ISSUE #1671: Important Guidelines on CFP forms are hardcoded
Hi!

We are hosting a conference on the FOSS United platform through our chapter, SciPy India, and have prepared a CFP form, which is available at https://fossunited.org/dashboard/cfp/apply/scipy-india/2026.

It seems that the "Important Guidelines" section is hardcoded in the form, and we don't have a way to replace or remove it with ours.

Image

This is a bit troublesome, as our guidelines contradict those mentioned in this section – we would like to link to a different proposal for an ideal proposal, we have our own Code of Conduct, and so on.

Could you please advise us on the best way to remove it, or could it be removed for our chapter?

Thank you!

cc: @Haleshot

idlip:
- fixes #1671 

- In general cfp Form we had cfp description shown on top and "important guidelines" hardcoded from global doctype to apply for all.

image

- Event Organizer might want to override or share other things, since we've a description text editor field we can simply reuse that to replace the global text.

- Removed form header since it was useless and having desc on all pages might be very long.
  Instead it is intended for proposers to read it once and proceed further simply

- If desc is empty then global guideline it shown anyway. If not any desc is shown as guidelines, so its upon organizer to write long form to inform proposer on any details and information.


idlip:
Merging as its small and easy fix to swap desc with guidelines if not empty.

Issue #1673, opened 2026-08-05
Some user-centric fields in the CFP form are not overrideable


agriyakhetarpal:
Hi! This is a similar issue to #1671:

<img width="373" height="101" alt="Image" src="https://github.com/user-attachments/assets/34da25d2-c53c-4b33-833d-1747bc18f658" />

<img width="577" height="190" alt="Image" src="https://github.com/user-attachments/assets/c207fe2d-f3a0-4bbc-a1af-81b2f02f7166" />

<img width="548" height="77" alt="Image" src="https://github.com/user-attachments/assets/84b7570a-ce65-4ec8-a2cc-d172bf84fca6" />

<img width="551" height="37" alt="Image" src="https://github.com/user-attachments/assets/2606144f-3927-4dc5-b2ae-e34e39d21928" />

<img width="543" height="163" alt="Image" src="https://github.com/user-attachments/assets/b73704e7-f59a-4991-b7e1-0038eb878efa" />

<img width="659" height="223" alt="Image" src="https://github.com/user-attachments/assets/0c838d6a-2fc5-49dc-8263-550a82ed5909" />

I am trying to run a test submission for our CFP, and I see the following issues:
- The Session Type is not overrideable
- The Session Category is not overrideable
- Nice to have: if we could remove the Contact Info field
- Regarding the newsletter line (fourth picture), it would be great if:
  - we could override the current line to point to our own newsletter, or website, or some other link such as a Code of Conduct information link
  - or configure removing it altogether
- The speaker image can be made optional if configurable, instead of mandatory
- The checkboxes at the end of the form (sixth picture) are not customisable

I hope y'all don't mind me describing all of these in one issue; we are trying to explore FOSS United as our platform and are running into these problems, and are in a bit of a time crunch lately as far as our announcements are concerned – and I figured it would be useful to discuss these as organisers of other events may run into them as well. Thank you!

cc: @malayajac @Haleshot @rahulporuri

agriyakhetarpal:
After making a test submission at https://fossunited.org/c/scipy-india/2026/cfp/5bs8jl6kkv, here are some things I noticed:
- Setting my picture as the speaker to an image also made it the profile picture for my account, which shouldn't happen. I set my older profile picture back through the dashboard.
[...]

PR #1727, opened 2026-08-28
feat(event-cfp): let organizers override session types, categories, and speaker fields per CFP

idlip:
## Status

- [x] Ready for review

---

## Related Issue

Closes / Addresses: #1673 

---

## Problem

Many fields in CFP form is catered for Indiafoss and generic FOSS events happening with fossunited.

We fail to provide customization for any event hosts to come and conduct an event in capable manner.

Firstly if cfp form can be extended from base default

---

## Solution

Keep default from fossunited as it is, but provide option to override and replace as they wish.

---

- [x] Tested locally
- [ ] Not adding much tests since these are just frontend UI helpers

---

## Changelog

<commit>: 
- doctype(event-cfp): add fields to override the cfpForm as per event
- helper(cfp): override session type and category as per event form
- add override helper to respective cfp edit, manage and proposal edit page
- doctype(event-cfp): validate session type match correctly to allowed fields

---

## Visualization

- Override limiting session type to 2 and category with only one and "other" for more manual
2026-08-28-155837_hyprshot

- final step showing only replaced checkbox from custom questions to agree
2026-08-28-155930_hyprshot

- Edit proposal page reflecting as per event CFP options to not expose others.
2026-08-28-160041_hyprshot


idlip:
ok, tested again end to end and works as expected.

unexpected always shows up in live prod when users use... ;)

Should be good to override.

TLDR;

By default everything is same, only text and things override when event cfp form changes and adds things.

PR #1740, opened 2026-09-02
ci: bump node 20 -> 24

idlip:
* just checking if it works and installs fine

- Adding bruno to package.json #1738  dev dependency complains on requiring node > 22..

frappe 16 needs > 24 only, so bumping to test if CI runs fine.


idlip:
CI runs.. lets go!!!

PR #1751, opened 2026-09-08
fix(free-ticket-code): avoid giving it to duplicate emails + avoid speaker coupon to mulitple counts

idlip:
- Auto speaker coupon incremented claim count with number of talk appearance they had in accepted proposal
  My bad for thinking it was appreciated to provide

  But now every speaker gets only one claim ticket for an event

- Avoid giving out coupon code for same email. This is to avoid giving out to same email for diversity and speaker if they co-exist.
- Just an helper, if need to be deliberate then organiser can use "+ticket" alias suffix to send.

This keeps it clean and avoid repeating for multiple external people.


idlip:
Should be good to go. It was minor controller change.

PR #1753, opened 2026-09-10
dashboard(eventDetails): add missing deck_link and schedule_description field

idlip:
- We missed both these fields in dashboard and until now were filling manually by DM request from organiser to add them.

- These are essential and missing gaps for conducting paid (ticket) events and dashboard should facilitate this.

- manage page showing deck link form field
image

- Sponsors manage page showing same for handy note
image

- Schedule page showing description text in plain markdown
image


idlip:
We'd do markdown preview in there itself, but bloat for one or two instance of use.
Skipping adding markdown preview with raw text box like github shows (code, preview)

Lets see if frappe-ui 1.0 has it for us

PR #1756, opened 2026-09-11
fix(frappe): frappe.request.path is not global safe exec point

idlip:
- Tried updated frpape v15 to latest thinking for security and fixes, but it removed frappe.request completely in jinja globals as safe exec.

- Had to fix it in ssh temporarily, but reported the issue and frappe contributor seem to have reverted for path 

https://github.com/frappe/frappe/pull/42748

Will keep this documented in case we want to update cloud and avoid getting that error.

To remove/revert after commit is backported to latest release

### DIFF
--- a/fossunited/fossunited/utils.py
+++ b/fossunited/fossunited/utils.py
@@ -45,6 +45,16 @@ def is_user_team_member(chapter, user):
     return False
 
 
+def get_request_path():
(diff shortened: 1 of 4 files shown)


idlip:
might need for deploy, and safe to have and revert with backport update.

PR #1767, opened 2026-09-16
fix(dev): make local setup self-bootstrapping

hanthor:
## Status

- [ ] Draft
- [x] Ready for review

---

## Related Issue

Fixes the local demo bootstrap workflow (no linked issue).

---

## Problem

`just demo` assumes a pre-existing Bench directory and hard-coded Docker Compose container names. A clean checkout fails before the site can be created, interrupted Bench initialization is mistaken for a valid install, the removed `patch_schema.py` is always invoked, and dashboard assets can be written to the wrong directory or fail silently when the app is linked from the workspace.

---

## Solution

Make the Justfile bootstrap and validate the complete local environment:

- Address services through Compose so Docker and Podman container naming differences do not matter.
- Create Bench when it is absent or incomplete, configure MariaDB and Redis, install required apps, and create or reset the site.
- Keep the workspace editable while using a small physical app shim for Frappe's relative asset imports.
- Build Vite output into the explicit app asset path, use Frappe's supported hard-link asset mode, and propagate esbuild failures.
- Run the legacy schema patch only on branches where the script exists.

---

- [x] `just --fmt --check`
- [x] `git diff --check`
- [x] Pre-commit hooks for `Justfile`, including Semgrep
- [x] Clean Bench initialization exercised locally
- [x] Existing Bench/site reset exercised locally
- [x] `just seed`
- [x] `just build-dashboard` including Vite and Frappe asset builds

---

## Changelog

`fix(dev): make local setup self-bootstrapping`

---

## Visualization

Not applicable; this changes the local development workflow.

---

## Further Ahead

None.


idlip:
I was simply using `fm` (frappe-manager) and wondered to try this.

and surprise it worked well. except for 2 change required.

Thank you for completing this again. works so well.

Now Imma see if anyone has complains on setting up our platform. Things should just work!
Need to once again check back for frappe-16 migration bump (#1589 )

Justfile
```diff
@@ -17,82 +17,165 @@ default:
...
+        if [ -L apps/fossunited ]; then
+            unlink apps/fossunited
+        elif [ -e apps/fossunited ] && [ ! -d apps/fossunited/fossunited ]; then
+            mv apps/fossunited "apps/fossunited.incomplete.$(date +%s)"
+        fi
+        mkdir -p apps/fossunited/fossunited
+        env/bin/pip install --quiet --editable /workspace
+        touch sites/apps.txt
+        grep -qx fossunited sites/apps.txt || echo fossunited >> sites/apps.txt
+        if [ ! -d apps/frappe_factory_bot ]; then
```
idlip: This appended fossunited to last line item itself.

```
frappe
frappe_factory_botfossunited
fossunited
```

it was like this for me and it errored for `setup`.
I think we can remove ? if frappe auto adds it via get-app install-app ?

idlip:
Also if you can add seed-data as a step so users can add it (ideally not required), just in case if everything is populated.

I got confused with default admin/password, thank you for mentioning it!

Justfile
```diff
@@ -17,82 +17,165 @@ default:
...
+            bench get-app --skip-assets --branch main https://github.com/harshtandiya/frappe_factory_bot
+        fi
```
idlip, line 107: Can you please link this to https://github.com/TomasBarry/frappe_factory_bot

It has a PR made by seed data person to support passing args or something.

idlip (changes requested):

Justfile
idlip: ModuleNotFoundError: No module named 'frappe_factory_botfossunited'

Justfile
```diff
@@ -17,82 +17,165 @@ default:
...
+setup: up
     #!/usr/bin/env bash
     set -euo pipefail
+    COMPOSE="{{ COMPOSE_CMD }}"
+    BENCH_DIR="/workspace/development/fossu-bench"
+
     echo "⏳ Waiting for MariaDB to be ready..."
-    until {{DOCKER}} exec devcontainer-mariadb-1 \
+    until $COMPOSE exec -T mariadb \
             mariadb -u root -p123 -e "SELECT 1" &>/dev/null; do
```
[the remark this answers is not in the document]
hanthor: Fixed — bounded the readiness loop to 60 retries (120s) and exit non-zero with a diagnostic if MariaDB never comes up.

Justfile
```diff
@@ -17,82 +17,165 @@ default:
...
-        bash -c "env/bin/pip install 'setuptools<80' --quiet"
-    {{DOCKER}} exec -w /workspace/development/fossu-bench devcontainer-frappe-1 bench --site fossunited.localhost install-app fossunited
-    {{DOCKER}} exec -w /workspace/development/fossu-bench devcontainer-frappe-1 bench --site fossunited.localhost migrate
-    # Patch: migrate doesn't add parent/parentfield/parenttype to FOSS Event CFP Review
-    # because it is both a standalone and a child doctype (see development/patch_schema.py).
-    {{DOCKER}} exec -w /workspace/development/fossu-bench/sites devcontainer-frappe-1 \
-        ../env/bin/python /workspace/development/patch_schema.py
+
+    if ! $COMPOSE exec -T frappe \
+            test -d "$BENCH_DIR/apps/frappe/frappe" -a -f "$BENCH_DIR/sites/common_site_config.json"; then
```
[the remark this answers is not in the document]
hanthor: Fixed — the reuse check now also requires `env/bin/pip` to be executable before skipping `bench init`.

Justfile
```diff
@@ -17,82 +17,165 @@ default:
...
+            --base=/assets/fossunited/dashboard/ \
+            --outDir=/workspace/fossunited/public/dashboard \
+            --emptyOutDir
+        cp /workspace/fossunited/public/dashboard/index.html \
+            /workspace/fossunited/www/dashboard.html
+    '
+    # Mirror public sources into a physical app path so Frappe's esbuild can resolve sibling apps.
+    $COMPOSE exec -T -u root -w /workspace/development/fossu-bench frappe bash -lc '
+        mkdir -p apps/fossunited/fossunited/public
+        cp -a /workspace/fossunited/public/. apps/fossunited/fossunited/public/
```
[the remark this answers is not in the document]
hanthor: Fixed — both `apps/fossunited/fossunited/public` and `apps/fossunited/node_modules/@knadh` are now `rm -rf`'d before the `cp -a`, so deleted source files no longer linger.

Justfile
hanthor: Confirmed the root cause: `echo fossunited >> sites/apps.txt` appended straight onto the previous line when that line lacked a trailing newline (e.g. left by `get-app`), producing `frappe_factory_botfossunited` and the `ModuleNotFoundError` you hit. Fixed by checking for a trailing newline and inserting one before appending if missing.

Justfile
hanthor: Checked — TomasBarry/frappe_factory_bot has the factory-override args support in an open, unmerged PR (TomasBarry/frappe_factory_bot#1), not on its `main` branch yet. Keeping `harshtandiya/frappe_factory_bot` for now; happy to switch once that PR lands upstream (or gets merged into the fork's main).

hanthor: Correction on my last reply — I checked further and the flags-passing feature is already present in the fork we use (`harshtandiya/frappe_factory_bot`), added upstream in TomasBarry's `main` back in Aug 2025 (`doctype.flags.update(overrides.get("flags", {}))` in `base_factory.py`), well before TomasBarry/frappe_factory_bot#1 was opened. `harshtandiya/frappe_factory_bot` is a fork of TomasBarry's repo and is in sync with that commit. So no change needed here — `StoreFactory.create(flags={...})` should already work with the current source. TomasBarry#1 is a small follow-up (pops `flags` out of the dict passed to `frappe.get_doc` instead of leaving it in) but isn't required for the feature itself. Let me know if you're hitting something specific that suggests otherwise.

hanthor: Actually tested this rather than just reasoning from history — ran it live against the bench our Justfile sets up:

```python
class ToDoFactory(BaseFactory):
    doctype = "ToDo"
    default_attributes = {"description": "..."}

doc = ToDoFactory.create(flags={"ignore_permissions": True, "custom_flag": "abc"})
```

Result: `"flags" not in doc.as_dict()` (doesnt leak into the saved doc), `doc.flags.ignore_permissions == True`, `doc.flags.custom_flag == "abc"`, and the full create -> insert -> delete cycle works. So flags-passing via factory overrides already works correctly with the current `harshtandiya/frappe_factory_bot` source — confirmed, not just inferred from commit history. No Justfile change needed.

Issue #1736, opened 2026-09-02
Proposals don't have OG info


rahulporuri:
See https://fossunited.org/c/indiafoss/2026/cfp/9lmiio7p5b for instance.

All of the necessary information exists already for us to show OG info

idlip:
Needs debugging and update on https://github.com/fossunited/ograph-gen

Even observed many proposal/profile page have red "X" mark

had tried once to replace that ograph-gen with svg somehow, but frappe cloud did not have libsvg install nor anyway to use that.

Gotta update that go inkscape generator itself for now.

idlip:
- https://fossunited.org/c/indiafoss/2026/cfp/62ntegtu09

^ preview here shows red "X" mark that looks bad

https://og.fossunited.org/gen/submission?talk_title=Building%20Open%20Source%20Digital%20Infrastructure%20[...]&session_type=Lightning%20Talk&event_name=IndiaFOSS%202026&speaker_designation=Overall%20Coordinators&speaker_name=DevCom%20IITB&speaker_image=/files/profile_dnme534lhd2bb918.webp&event_chapter=INDIAFOSS

![](https://og.fossunited.org/gen/submission?talk_title=Building%20Open%20Source%20Digital%20Infrastructure%20[...]&session_type=Lightning%20Talk&event_name=IndiaFOSS%202026&speaker_designation=Overall%20Coordinators&speaker_name=DevCom%20IITB&speaker_image=/files/profile_dnme534lhd2bb918.webp&event_chapter=INDIAFOSS)

idlip:
1. For first one

![](https://og.fossunited.org/gen/submission?talk_title=From%20DPI%20to%20People%E2%80%99s%20Digital%20Infrastructure:%20[...]&session_type=Panel%20Discussion&event_name=IndiaFOSS%202026&speaker_designation=SAFAR&speaker_name=Sree%20Harsha%20Tanneru&speaker_image=None&event_chapter=INDIAFOSS)

2. ![](https://og.fossunited.org/gen/submission?talk_title=Building%20Open%20Source%20Digital%20Infrastructure%20[...]&session_type=Lightning%20Talk&event_name=IndiaFOSS%202026&speaker_designation=Overall%20Coordinators&speaker_name=DevCom%20IITB&speaker_image=/files/profile_dnme534lhd2bb918.webp&event_chapter=INDIAFOSS)

sidd190:
> 1. For first one
> 
[...]

PR #1790, opened 2026-09-22
fix: wire up OG image tags for profile and RSVP pages

sidd190:
## Summary

- **Profile**: fix param names sent to ograph-gen (`full_name`→`profile_name`, `image`→`profile_image`) so the `/gen/profile` endpoint renders correctly
- **RSVP**: replace hardcoded generic `og-image.png` with dynamic `/gen/rsvp` endpoint showing event chapter and name, with fallback if Ograph Settings isn't configured

Ref: #1736


idlip:
Please update description and add visuals for all cases?

